PR #6, opened 2023-08-30
Initial checkin for OFM frontend 1st iteration scaffolding. Common co…

jgstorey:
Common code was copied from STUDENT-ADMIN.


weskubo-cgi (approved):
Initial commit approved.
There is still some pending work around component structure and renaming.

PR #7, opened 2023-09-05
Commiting intital OFM backend code based on MyCCS backend.

jgstorey:
I left the following files in the backend root assuming we'll need them or need to update their contents with config specific to OFM (i.e. .babelrc, Dockerfile, sonar-project.properties).


weskubo-cgi (approved):
Approved. Project files will be refactored/removed once development starts.

PR #13, opened 2023-09-25
Js fix console errs dynamic env params

jgstorey:
Renamed branch... removed & in name.


frontend/vite.config.js
```diff
@@ -19,5 +19,16 @@ export default defineConfig({
...
+
+
+
+
+
+
+
+
+)
+
```
weskubo-cgi, line 34: I'm not sure why this change is here. Seems to be mostly formatting and empty lines.

frontend/src/router/index.js
```diff
@@ -170,7 +170,7 @@ const router = createRouter({
 router.beforeEach((to, _from, next) => {
   // this section is to set page title in vue store
   const appStore = useAppStore()
-  if (to && to.meta) {
+  if (to && to.meta.pageTitle) {
```
weskubo-cgi, line 173: I think it might make more sense to leave as it but ensure that Login route has a pageTitle defined.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+    "configurations": [
+        {
+            "type": "msedge",
+            "request": "launch",
+            "name": "Launch Edge against localhost",
+            "url": "http://localhost:8082",
+            "webRoot": "${workspaceFolder}/frontend/src/*"
+        }
+    ]
+}
```
weskubo-cgi, line 15: I don't think we want this file in git. It's accounted for the in the .gitignore in backend and frontend but I think we will need a top level .gitignore as well. Just add the entry for this to start (you can copy from existing files).

weskubo-cgi (changes requested):

frontend/src/router/index.js
jgstorey: Note: adding page title to the check removes console warning... 
image

.vscode/launch.json
jgstorey: Fixed.

frontend/vite.config.js
jgstorey: Fixed.

weskubo-cgi (approved):

PR #29, opened 2023-11-03
Updated so when bceid singin, the username is sent to the backend

jgstorey:


backend/src/components/user.js
```diff
@@ -76,9 +76,9 @@ async function getUserInfo(req, res) {
       return res.status(HttpStatus.OK).json(resData)
     }
   } else {
-    //Not an idir user, so just get the guid from the header
-    log.verbose('User Guid is: ', userGuid)
-    userResponse = await getUserProfile(userGuid, null)
+    // BCeID user
```
weskubo-cgi, line 79: Can we add a comment to explain what's happening here in terms of GUID population in Dynamics?

weskubo-cgi (changes requested):

backend/src/components/user.js
jgstorey: Updated comment.

weskubo-cgi (approved):

PR #31, opened 2023-11-06
Qafixes ofmcc 251 view notifications

jgstorey:
These changes fix the original issues reported from QA, as well improves upon the UI in some small areas (i.e. read/unread buttons and headers).
FYIs:
- I found a new bug that can occur when updating sorted data.
- You'll notice i increased the vue and vuetify versions...  as per my mail on the subject but i rolled that back in last commit as my changes still work. Still think we should upgrade though based on the fact some vuetify doc examples do not work with our versions.
- On MessagingVue I needed to change the font size on v-badge/v-tab to make it position reasonably for both single and double digit unread counts. I could not find a way to do it without using the 'deep selector', applying a simple style would not work. if 3 digits unread count the v-badge will spill into next tab abit... left that fine tuning as a nice to have.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -21,10 +21,10 @@
         "sass": "^1.66.1",
         "sass-loader": "^13.3.2",
         "vite-plugin-vuetify": "^1.0.2",
-        "vue": "^3.3.7",
+        "vue": "^3.3.4",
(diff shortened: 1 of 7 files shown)


weskubo-cgi (approved):
Approved.

PR #34, opened 2023-11-07
Qafixes ofmcc 251 view notifications

jgstorey:
Fixed:
- updating read/unread when list is sorted retains sort and updates expected notices in backend.
- minor update to notifications indicator on notifications tab.

### DIFF
--- a/frontend/src/components/notifications/NotificationsTab.vue
+++ b/frontend/src/components/notifications/NotificationsTab.vue
@@ -164,8 +164,8 @@ export default {
      * Update the body/item checkboxes to read or unread.
      */
     updateBodyCheckboxesReadUnread(isRead) {
-      const areAllNotificatoinsChecked = this.bodyCheckboxesSelected.every((item) => item === true)
-      if (areAllNotificatoinsChecked) {
+      const areAllNotificationsChecked = this.bodyCheckboxesSelected.every((item) => item === true)
+      if (areAllNotificationsChecked) {
         this.notifications.forEach((notification) => {
           notification.isRead = isRead
           this.updateNotification(notification)
@@ -225,7 +225,7 @@ hr {
 }
 
 .btn-style {
-  padding: 0px 6px !important;
+  padding: 0px 6px;
   margin: 0px;
   font-size: 14px;
   background-color: #ffffff;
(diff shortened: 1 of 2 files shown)


frontend/src/components/notifications/NotificationsTab.vue
```diff
@@ -152,19 +151,32 @@ export default {
...
      * Update the body/item checkboxes to read or unread.
      */
     updateBodyCheckboxesReadUnread(isRead) {
-      this.bodyCheckboxesSelected.forEach((item, index) => {
-        if (item) { // jstorey can remove?
-          this.notifications[index].isRead = isRead
-          this.updateNotification(this.notifications[index], isRead)
-        }
-      })
+      const areAllNotificatoinsChecked = this.bodyCheckboxesSelected.every((item) => item === true)
```
weskubo-cgi, line 167: Typo.

frontend/src/components/notifications/NotificationsTab.vue
```diff
@@ -260,6 +272,10 @@ hr {
   font-weight: bold;
 }
 
+.subject-text {
+  font-size: 18px !important;
```
weskubo-cgi, line 276: !important should be avoided in css. If it's the only way it works for now we can revisit.

frontend/src/views/MessagingView.vue
```diff
@@ -90,5 +90,7 @@ export default {
 
 :deep(.v-badge__badge) {
   font-size: 0.75rem !important;
+  min-width: 18px !important;
+  padding: 2px 3px !important;
```
weskubo-cgi, line 94: !important should be avoided in css. If it's the only way it works for now we can revisit.

weskubo-cgi (changes requested):

frontend/src/components/notifications/NotificationsTab.vue
jgstorey: Fixed.

frontend/src/components/notifications/NotificationsTab.vue
jgstorey: Fixed

frontend/src/views/MessagingView.vue
jgstorey: Fixed, removed all !import and ensured styles still take effect. Makes me feel crazy... i only added !important because i wasn't seeing the styles take effect in the 1st place.

weskubo-cgi (approved):

PR #35, opened 2023-11-07
Add ability to switch facilities

weskubo-cgi:
- Added new facility header
- Refactored facilities in auth store
- Removed some usage of localStore in auth store
- Added some empty views/routes to complete menu


jgstorey (approved):
Review confirmed.

PR #40, opened 2023-11-16
OFMCC 380 - list requests, & OFMCC 589 - display portal status base on CRM status

vietle-cgi:
OFMCC 380 - list requests & OFMCC 589 - display portal status base on CRM status

### DIFF
--- a/frontend/src/components/messages/AssistanceRequestTable.vue
+++ b/frontend/src/components/messages/AssistanceRequestTable.vue
@@ -30,7 +30,7 @@
     </template>
     <template #item.lastConversationTime="{ item }">
       <div :class="getItemClass(item)">
-        {{ convertDateToLocalDate(item.lastConversationTime) }}
+        {{ formatDate(item.lastConversationTime) }}
       </div>
     </template>
   </v-data-table-virtual>
@@ -39,6 +39,7 @@
 <script>
 import { mapState, mapActions } from 'pinia'
 import { useMessagesStore } from '@/stores/messages'
+import moment from 'moment'
 
 export default {
   name: 'AssistanceRequestTable',
@@ -132,41 +133,20 @@ export default {
       return item?.status === 'Action required'
     },
     getItemClass(item) {
-      let result = ''
-      if (this.isActionRequiredMessage(item)) {
-        result += 'action-required-message '
+      return {
+        'action-required-message': this.isActionRequiredMessage(item),
+        'unread-message': !item?.isRead,
+        'highlighted-row': this.selectedRequestId === item?.assistanceRequestId,
       }
-      if (!item?.isRead) {
-        result += 'unread-message '
-      }
-      if (this.selectedRequestId === item?.assistanceRequestId) {
-        result += 'highlighted-row '
-      }
-      return result
     },
-    convertDateToLocalDate(date) {
-      let localDate = new Date(date)
-      let monthNames = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
-      let dateString = localDate?.getFullYear() + '-' + monthNames[localDate?.getMonth()] + '-' + localDate?.getDate()
-      return dateString
+    formatDate(date) {
+      return moment(date).format('YYYY-MMM-DD')
     },
   },
 }
 </script>
 
 <style scoped>
-th {
-  padding: 0px 0px 0px 4px !important;
-}
-td {
-  padding: 0px 3px 0px 4px !important;
-}
-hr {
-  border: 0;
-  height: 1px;
-  background: #6699cc;
-  background-image: linear-gradient(to right, #6699cc, #6699cc, #6699cc);
-}
 .tableText {
   font-size: small;
 }
@@ -184,7 +164,7 @@ hr {
 .highlighted-row {
   display: flex;
   align-items: center;
-  background: #d4eaff !important;
+  background: #d4eaff;
   height: 100%;
 }
 :deep(.v-data-table__td) {
--- a/frontend/src/components/messages/AssistanceRequestTable.vue
+++ b/frontend/src/components/messages/AssistanceRequestTable.vue
@@ -10,50 +10,28 @@
     show-select
     hover
     fixed-header
-    density="compact">
-    <!-- HEADERS -->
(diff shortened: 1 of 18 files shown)


frontend/src/components/messages/AssistanceRequestTable.vue
```diff
@@ -0,0 +1,193 @@
...
+    </template>
+  </v-data-table-virtual>
+</template>
+
+<script>
+import { mapState, mapActions } from 'pinia'
+import { useMessagesStore } from '@/stores/messages'
+
+export default {
+  name: 'AssistanceRequestTable',
```
weskubo-cgi, line 44: Not a big deal but try to follow a standard order for components:

https://eccbc.atlassian.net/wiki/spaces/OOFMCC/pages/152698987/OFMCC+-+Portal+Home

frontend/src/components/messages/AssistanceRequestTable.vue
```diff
@@ -0,0 +1,193 @@
...
+        this.selectedAssistanceRequest.isRead = isRead
+        payload.isRead = isRead
+      }
+      await this.updateAssistanceRequest(assistanceRequestId, payload)
+    },
+    isActionRequiredMessage(item) {
+      return item?.status === 'Action required'
+    },
+    getItemClass(item) {
+      let result = ''
```
weskubo-cgi, line 135: You might find it useful to do object style class binding instead: https://vuejs.org/guide/essentials/class-and-style.html rather than string concatentation.

frontend/src/components/messages/AssistanceRequestTable.vue
```diff
@@ -0,0 +1,193 @@
...
+        result += 'highlighted-row '
+      }
+      return result
+    },
+    convertDateToLocalDate(date) {
+      let localDate = new Date(date)
+      let monthNames = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
+      let dateString = localDate?.getFullYear() + '-' + monthNames[localDate?.getMonth()] + '-' + localDate?.getDate()
+      return dateString
+    },
```
weskubo-cgi, line 152: Can we bring in moment() or another library to do date formatting/parsing? Also this might be best in utils if there's a lot of code. Though if it's just a simple in moment then we can leave it here.

frontend/src/components/messages/AssistanceRequestTable.vue
```diff
@@ -0,0 +1,193 @@
...
+</script>
+
+<style scoped>
+th {
+  padding: 0px 0px 0px 4px !important;
+}
+td {
+  padding: 0px 3px 0px 4px !important;
+}
+hr {
```
weskubo-cgi, line 164: I don't see an hr in the template.

frontend/src/components/messages/AssistanceRequestTable.vue
```diff
@@ -0,0 +1,193 @@
...
+}
+</script>
+
+<style scoped>
+th {
+  padding: 0px 0px 0px 4px !important;
+}
+td {
+  padding: 0px 3px 0px 4px !important;
+}
```
weskubo-cgi, line 163: It doesn't seem that these actually have any effect but if required can we a) use class names and b) avoid !important if possible.

frontend/src/components/messages/MessagesTab.vue
```diff
@@ -1,136 +1,104 @@
...
+              </v-btn>
+              <v-btn class="btn-style" @click="toggleMarkReadButton()">
+                <v-icon class="icon" left>mdi-email-open-outline</v-icon>
+                <span class="btn-label">Mark read</span>
+              </v-btn>
+            </div>
+          </v-col>
+        </v-row>
+        <v-row v-if="!assistanceRequests" class="mx-1">
+          <v-skeleton-loader type="table-tbody"></v-skeleton-loader>
```
weskubo-cgi, line 24: Not sure if it will work for you but I think it's normal to embed the content in the skeleton-loader itself. You could probably bind the loading property to assistanceRequests. See https://vuetifyjs.com/en/components/skeleton-loaders/#loading.

weskubo-cgi (changes requested):

frontend/src/components/messages/AssistanceRequestTable.vue
vietle-cgi: when I followed the orders in the Confluence page, I got the Eslints warnings (i.e.: The "props" property should be above the "data" property and The "watch" property should be above the "methods" property).

weskubo-cgi: Ok. We can revisit our standards and/or linter config.

frontend/src/stores/messages.js
```diff
@@ -1,74 +1,69 @@
...
+        if (assistanceRequestId && !isEmpty(payload)) {
+          await ApiService.apiAxios.put(ApiRoutes.MESSAGE + '/' + assistanceRequestId, payload)
+        }
       } catch (error) {
-        console.log(`Failed to create a new Assistance Request - ${error}`)
+        console.log(`Failed to update existing assistance request - ${error}`)
         throw error
       }
     },
+    async addNewAssistanceRequestToStore(assistanceRequestId) {
```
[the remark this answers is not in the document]
vietle-cgi: I've done it this way, so the user won't need to wait for our app to reload all messages. That reloading can take long time if there are lots of messages. If you prefer to reload all messages instead, I will update it.

frontend/src/stores/messages.js
```diff
@@ -1,74 +1,69 @@
 import { ApiRoutes } from '@/utils/constants'
 import ApiService from '@/common/apiService'
 import { defineStore } from 'pinia'
+import { isEmpty } from 'lodash'
 
 export const useMessagesStore = defineStore('messages', {
   namespaced: true,
   state: () => ({
-    messages: null,
+    assistanceRequests: null,
```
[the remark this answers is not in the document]
vietle-cgi: Agreed! This naming inconsistency is confusing. I chose "assistanceRequests" at the beginning because I followed the naming convention in CRM. We can make this change in the future

frontend/src/stores/messages.js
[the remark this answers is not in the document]
weskubo-cgi: Nope. That's fine. Thanks for explaining.

weskubo-cgi (approved):

PR #42, opened 2023-11-20
Ofmcc 636 setup uat environment

SoLetsDev:


.github/workflows/deploy-to-openshift-backend-dev.yml
```diff
@@ -9,6 +9,8 @@ env:
   OPENSHIFT_TOKEN: ${{ secrets.OPENSHIFT_TOKEN }}
   # 🖊️ EDIT to set the kube context's namespace after login. Leave blank to use your user's default namespace.
   OPENSHIFT_NAMESPACE: ${{ secrets.OFM_NAMESPACE_NO_ENV }}-dev
+  # We have multiple environments in each openshift namespace dev, test in DEV and uat, efx in TEST. This setting will help name the deployment config and routes for each.
+  OPENSHIFT_ENVIRONMENT: 'dev'
```
weskubo-cgi, line 13: Re: your comment in chat maybe we can just go with ENVIRONMENT? I don't think the concept of environment is unique to OpenShift as it also applies GitHub environments and it is the terminology we generally use. I'm not too bothered either way though.

Also we can reference this below:

  openshift-ci-cd:
    name: Deploy Backend to TEST
    runs-on: ubuntu-latest
    environment: **test**

weskubo-cgi (changes requested):

.github/workflows/deploy-to-openshift-backend-dev.yml
SoLetsDev: updated to use APP_ENVIRONMENT

weskubo-cgi (approved):

PR #48, opened 2023-11-28
Ofmcc 658 add request single facility

weskubo-cgi:
- Default facility to current facility
 - Added Select All capability
 - Switched facility select to use chips
 - Added standard modal label


jgstorey (approved):
Reviewed.

SoLetsDev (approved):

PR #51, opened 2023-12-02
OFMCC-728 QA fixes: filter out facilities with no portal access. Colu…

jgstorey:
…mn label change. Fix to default sort.

### DIFF
--- a/frontend/src/views/SettingsView.vue
+++ b/frontend/src/views/SettingsView.vue
@@ -26,7 +26,7 @@
       <v-col>
         <!-- Users Table -->
         <v-skeleton-loader :loading="!usersAndFacilities" type="table-tbody">
-          <v-data-table :headers="headersUsers" :items="sort(filteredUserFacilities)" item-key="contactId" item-value="contactId" show-expand density="compact" :expanded.sync="expanded">
+          <v-data-table :headers="headersUsers" :items="filteredUserFacilities" item-key="contactId" item-value="contactId" show-expand density="compact" :expanded.sync="expanded">
             <!-- Slot to customize expand row event -->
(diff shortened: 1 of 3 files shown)


frontend/src/views/SettingsView.vue
```diff
@@ -26,7 +26,7 @@
       <v-col>
         <!-- Users Table -->
         <v-skeleton-loader :loading="!usersAndFacilities" type="table-tbody">
-          <v-data-table :headers="headersUsers" :items="filteredUserFacilities" item-key="contactId" item-value="contactId" show-expand density="compact" :expanded.sync="expanded">
+          <v-data-table :headers="headersUsers" :items="sort(filteredUserFacilities)" item-key="contactId" item-value="contactId" show-expand density="compact" :expanded.sync="expanded">
```
weskubo-cgi, line 29: Would it make more sense to just do the sort in the computed filteredUserFacilities method? Also, in filteredUserFacilities it might make sense to return early (with sort) if there's no facilityNameFilter defined. I assume now it's checking for contains '' and the result ends up true which isn't necessarily intuitive.

frontend/src/views/SettingsView.vue
```diff
@@ -201,6 +203,28 @@ export default {
           return 'Inactive'
       }
     },
+
+
+    /**
+     * Sort users by: account management role 1st, expense authority 2nd, last name 3rd
+     */
+    sort(users) {
```
weskubo-cgi, line 211: Can we be more specific with this name? E.g. sortUsers(). In theory we could have multiple sort methods.

jgstorey: Fixed.

weskubo-cgi (approved):

PR #52, opened 2023-12-05
ofmcc-405 - update getNotifications query to only include emails with statecode equals to 1 (Completed)

vietle-cgi:
Update getNotifications query to only include emails with statecode equals to 1 (Completed)


jgstorey (approved):
Review confirmed.

PR #59, opened 2023-12-14
ofmcc-1001 - add pageSize to GET Notifications request

vietle-cgi:
Add pageSize to GET Notifications to raise the max number of records in GET response payload (new limit = 500 records)


jgstorey (approved):
Reviewed.

PR #63, opened 2023-12-19
Ofmcc 729 edit user

jgstorey:


backend/src/components/user.js
```diff
@@ -203,10 +205,13 @@ function mapUserFacilityObjectForFront(data) {
...
 }
 
-async function getUserFacilities(req, res) {
+async function getUserFacilities(req, res, onlyWithPortalAccess) {
   try {
     let userFacilities = []
-    const operation = `ofm_bceid_facilities?$expand=ofm_facility($select=accountnumber,address1_composite,name)&$filter=(statecode eq 0 and _ofm_bceid_value eq ${req.params.contactId}) and (ofm_facility/statecode eq 0)`
+    let operation = `ofm_bceid_facilities?$expand=ofm_facility($select=accountnumber,address1_composite,name)&$filter=(statecode eq 0 and _ofm_bceid_value eq ${req.params.contactId}) and (ofm_facility/statecode eq 0)`
+    if (onlyWithPortalAccess) {
+      operation = operation + ` and (ofm_portal_access eq true)`
```
weskubo-cgi, line 213: Technically this doesn't have to be a template literal since there's no variables but it's also fine as is (and consistent with previous strings).

backend/src/components/user.js
```diff
@@ -203,10 +205,13 @@ function mapUserFacilityObjectForFront(data) {
   return userFacilities
 }
 
-async function getUserFacilities(req, res) {
+async function getUserFacilities(req, res, onlyWithPortalAccess) {
   try {
     let userFacilities = []
```
weskubo-cgi, line 210: Keep an eye on let/const usage and prefer const wherever possible knowing that const can be mutated just not reassigned.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
 async function createUser(req, res) {
   try {
     const payload = mapUserObjectForBack(req.body)
-    let response = await postOperation('contacts', JSON.stringify(payload))
+    let response = await postOperation('contacts', payload)
+    let returnVal = new MappableObjectForFront(response, UserMappings).toJSON()
```
weskubo-cgi, line 234: These can both be const.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
...
+    returnVal.role = Number(returnVal.role)
+    return res.status(HttpStatus.OK).json(returnVal)
+  } catch (e) {
+    return res.status(HttpStatus.INTERNAL_SERVER_ERROR).json(e.data ? e.data : e?.status)
+  }
+}
+
+function mapContactForBack(data) {
+  let contact = new MappableObjectForBack(data, ContactMappings).toJSON()
```
weskubo-cgi, line 243: This can be const.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
...
+  contact.entityNameSet = 'contacts'
+  contact.actionMode = 'Update'
+  return contact
+}
+
+function mapContactFacilitiesForBack(data) {
+  let facilitiesForBack = []
```
weskubo-cgi, line 250: Check let/const usage in function.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
...
+    facility.entityID = facilityFromFront.bceidFacilityId
+    facility.ofm_portal_access = facilityFromFront.ofmPortalAccess
+    facility.entityNameSet = 'ofm_bceid_facilities'
+    facility.actionMode = 'Update'
+    facilitiesForBack.push(facility)
+  })
+  return facilitiesForBack
+}
+
+async function updateUser(req, res) {
```
weskubo-cgi, line 262: Check let/const usage in function.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
...
+  try {
+    let payload = {
+      batchTypeId: 101,
+      feature: 'AccountManagement',
+      function: 'UserEdit',
+      actionMode: 'Update',
+      scope: 'Parent-Child',
```
weskubo-cgi, line 269: In the future if we're seeing these values used in multiple places we may want to make constants for some of these.

backend/src/components/user.js
```diff
@@ -225,7 +230,57 @@ function mapUserObjectForBack(data) {
...
+        ofm_bceid_facility: [],
+      },
+    }
+    const updatedContact = mapContactForBack(req.body)
+    const updatedFacilities = mapContactFacilitiesForBack(req.body)
+    log.info('updatedContact updatedContact:', JSON.stringify(updatedContact, null, 2))
+    payload.data.contact = updatedContact
+    log.info('contact type:', typeof updatedContact)
+    payload.data.ofm_bceid_facility = updatedFacilities
+    log.info('facilities type:', typeof updatedFacilities)
```
weskubo-cgi, line 281: Can we remove log statements?
I wouldn't actually bother defining a const for these since all we do is assign it to the payload. My general rule is don't define a variable if it's only referenced once unless it significantly improves readability.

backend/src/components/utils.js
```diff
@@ -251,6 +251,27 @@ function errorResponse(res, msg, code) {
   })
 }
 
+async function postBatches(payload) {
```
weskubo-cgi, line 254: Can we clean up logging here?

backend/src/components/utils.js
```diff
@@ -251,6 +251,27 @@ function errorResponse(res, msg, code) {
...
+  const url = config.get('dynamicsApi:apiEndpoint') + '/api/batches'
+  log.info('postBatches Url', url)
+
+  if (log.isInfoEnabled) {
```
weskubo-cgi, line 258: Note this checks isInfoEnabled but then logs at verbose level.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -214,7 +213,10 @@ export default {
           if (this.isAddingUser) {
```
weskubo-cgi, line 213: Minor point but on line 211 I would do an early return if not complete as it avoids having all your logic nested in the if block.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -240,17 +244,27 @@ export default {
     /**
      * Update user and emit success/fail event.
      */
-    async updateUser() {
+    async updateUser(user) {
       try {
-        //TODO - complete when API is ready
-        //this.user.organizationId = this.userInfo.organizationId
-        //const response = await ApiService.apiAxios.post(ApiRoutes.USER + '/update', this.user)
+        const response = await ApiService.apiAxios.post(ApiRoutes.USER + '/update', this.user)
```
weskubo-cgi, line 249: Response is declared but not use. Your linter should pick this up.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -230,6 +232,8 @@ export default {
       try {
         this.user.organizationId = this.userInfo.organizationId
         const response = await ApiService.apiAxios.post(ApiRoutes.USER + '/create', this.user)
```
weskubo-cgi, line 234: Not a problem for now but we should probably revisit some of our endpoints as they don't follow REST standards. I'm curious how things are in CCOF.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -240,17 +244,27 @@ export default {
...
         this.$emit('update-success-event', false, error)
       }
     },
 
+    /**
+     * Get the list of facilities by contact id. If onlyWithPortalAccess is true, only return facilities with portal access.
+     */
+    async getUserFacilities(contactId, onlyWithPortalAccess) {
+      try {
+        const res = await ApiService.apiAxios.get(ApiRoutes.USER_FACILITIES + '/' + onlyWithPortalAccess + '/' + contactId)
```
weskubo-cgi, line 261: I would expect a URL of this sort to be something like:

/users/:contactId/facilities?onlyWithPortalAccess=true|false

It works for now but we should probably clean it up later.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -274,6 +288,82 @@ export default {
...
       return ADD_USER_SUCCESS_MSG
     },
+
+    /**
+     * Returns true if users facility access has changed, false otherwise.
+     */
+    hasUserFacilityAccessChanged(selectedFacilityIds, userFacilities) {
+      if (selectedFacilityIds?.length !== userFacilities?.length) {
+        return true;
+      } else {
```
weskubo-cgi, line 298: There's no need for an else block since the if block returns.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -274,6 +288,82 @@ export default {
...
+        for (let facilityId of selectedFacilityIds) {
```
weskubo-cgi, line 299: This can be const.

weskubo-cgi, line 299: I think this loop can be simplified. Something like https://stackoverflow.com/questions/16312528/check-if-an-array-contains-any-element-of-another-array-in-javascript.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -274,6 +288,82 @@ export default {
...
+     * Get updated user facility access.
+     */
+    async getUpdatedFacilityAccess(user, selectedFacilityIds) {
+      let facilitiesToAdd = []
+      let facilitiesToRemove = []
+      // Get facility objects selected by facilityIds
+      const selectedFacilities = this.getSelectedFacilitiesByIds(selectedFacilityIds)
+      // Get users current facilities
+      const userFacilities = await this.getUserFacilities(user.contactId, false)
+      if (Object.keys(user.facilities).length === 0 && user.facilities.constructor === Object) {
```
weskubo-cgi, line 318: I'm not too sure what this constructor check does.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -80,11 +80,11 @@
             </v-col>
```
weskubo-cgi, line 80: On line 3 the min-width can be removed as it's now in AppDialog.

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -151,7 +150,10 @@ export default {
       rules,
       isDisplayed: false,
       isLoading: false,
-      facilities: [],
+      user: {},
+      selectedFacilityIds: [],
+      facilitiesToAdminister: [],
+      facilitiesUser: [],
       userOperationType: '', // Can be 'adding', 'updating'
```
weskubo-cgi, line 157: If there's no precedent can we just use 'add'/'update'?

frontend/src/components/settings/ManageUserDialog.vue
```diff
@@ -175,19 +177,16 @@ export default {
         this.isDisplayed = value
       },
     },
-    user: {
-      handler() {
-        this.userOperationType = Object.keys(this.user).length === 0 ? 'adding' : 'updating';
+    updatingUser: {
+      handler(value) {
+        this.userOperationType = Object.keys(value).length === 0 ? 'adding' : 'updating';
```
weskubo-cgi, line 182: I might be inclined to have the operation type as a prop rather than use a rule of thumb to determine the operation (e.g. if you wanted to default a value in the user that's passed in it wouldn't be empty). But I don't think it's a necessary change for now.

weskubo-cgi (changes requested):

backend/src/components/user.js
jgstorey: Fixed.

backend/src/components/user.js
jgstorey: Fixed.

backend/src/components/user.js
jgstorey: Fixed.

backend/src/components/user.js
jgstorey: Understood. Left as is.

backend/src/components/user.js
jgstorey: Fixed.

backend/src/components/utils.js
jgstorey: The logging in postBatches is consistent with many other similar methods in utils following the same logging pattern... the logging appears helpful torwards log and debugging. Please confirm if you still want removed.

backend/src/components/utils.js
jgstorey: Consistent with existing logging patterns in component. Please confirm you want changed.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Fixed

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Fixed.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Fixed.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Fixed.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: checks facilities is an object.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Fixed.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Updated.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: Understood. Left for now.

backend/src/components/utils.js
weskubo-cgi: Leave for now.

backend/src/components/utils.js
weskubo-cgi: Leave for now.

frontend/src/components/settings/ManageUserDialog.vue
weskubo-cgi: To review with team.

backend/src/components/user.js
weskubo-cgi: payload and response are still let.

frontend/src/components/settings/ManageUserDialog.vue
weskubo-cgi: Were you able to look at a simpler solution?

backend/src/components/user.js
jgstorey: Fixed for real this time.

frontend/src/components/settings/ManageUserDialog.vue
jgstorey: sorry, missed this... addressed with .some .includes.

weskubo-cgi (approved):

PR #73, opened 2024-01-09
QA fixes: added phone, cell, and email to facility, and inactive lice…

jgstorey:
…nces will no longer appear in facility account management.


backend/src/components/facilities.js
```diff
@@ -28,7 +28,7 @@ async function getFacilityContacts(req, res) {
 
 async function getFacilityLicences(req, res) {
   try {
-    const operation = `ofm_licences?$select=ofm_health_authority,ofm_licence,ofm_licenceid,statuscode&$filter=(_ofm_facility_value eq ${req.params.facilityId})`
+    const operation = `ofm_licences?$select=ofm_health_authority,ofm_licence,ofm_licenceid,statuscode&$filter=(_ofm_facility_value eq ${req.params.facilityId}) and (statuscode eq 1)`
```
vietle-cgi, line 31: I am not sure if in the future there will be more status codes for this one and they want to display them as well? I normally use statecode if I want to decide an entity is Active or Inactive.

jgstorey: Found instances of existing use of statecode on filters, will push through for sake of sprint and ask Hoang's confirmation if these need to change to statecode.

vietle-cgi (approved):

PR #75, opened 2024-01-10
adding select all for facilities when managing users

SoLetsDev:


frontend/src/components/account-mgmt/ManageUserDialog.vue
```diff
@@ -169,6 +182,12 @@ export default {
     isSameUser() {
       return this.user.userName === this.userInfo.userName
     },
+    allFacilitiesSelected() {
+      return this.selectedFacilityIds.length === this.facilitiesToAdminister.length
+    },
+    someFacilitiesSelected() {
+      return this.selectedFacilityIds.length !== this.facilitiesToAdminister.length
```
vietle-cgi, line 189: With the current logic, I think someFacilitiesSelected() is same as **!**allFacilitiesSelected(). Do we need to have both of them?

vietle-cgi: FYI, in NewRequestDialog.vue, someFacilitiesSelected() checks if the array's length > 0

SoLetsDev: copied the logic from newRequestDialog.

PR #91, opened 2024-02-01
ofmcc-1131 - filter out inactive licences using statecode instead of statuscode

vietle-cgi:


jgstorey (approved):
Reviewed and approved.

PR #106, opened 2024-02-21
ofmcc-1161 - Application Review page + route leave behavior enhancement

vietle-cgi:


frontend/src/components/applications/review/YearlyFacilityCostSummary.vue
```diff
@@ -0,0 +1,152 @@
...
+import { FACILITY_TYPES } from '@/utils/constants'
+import { formatDecimalNumber } from '@/utils/common'
+
+export default {
+  components: { AppLabel },
+  computed: {
+    ...mapState(useApplicationsStore, ['currentApplication']),
+  },
+  created() {
+    this.facilityType = this.currentApplication?.facilityType
```
weskubo-cgi, line 140: I think this probably more appropriate as a computed value as it's technically reactive.

weskubo-cgi (changes requested):

frontend/src/components/applications/review/YearlyFacilityCostSummary.vue
vietle-cgi: it is updated.

weskubo-cgi (approved):

PR #107, opened 2024-02-22
ofmcc-16 - application submit and confirmation

vietle-cgi:


frontend/src/router/index.js
```diff
@@ -174,11 +176,24 @@ const router = createRouter({
...
+          name: 'application-confirmation',
+          component: ApplicationConfirmationView,
+        },
       ],
       meta: {
         requiresAuth: true,
       },
     },
+    {
+      path: '/supp-applications',
```
weskubo-cgi, line 190: I think the terminology within the portal is Supplementary Allowance. So maybe /supp-allowances and SupplementaryAllowanceView?

weskubo-cgi (changes requested):

weskubo-cgi (approved):

frontend/src/router/index.js
vietle-cgi: it was fixed.

weskubo-cgi (approved):

PR #113, opened 2024-02-29
Ofmcc 1594 add inclusion to org

jgstorey:


frontend/src/components/organizations/OrganizationInfo.vue
```diff
@@ -216,55 +221,95 @@ export default {
...
       },
     },
-    loading: {
+    uploadedDocuments: {
+      type: Array,
+      default: () => {
+        return []
+      },
+    },
+    loadingAll: {
```
weskubo-cgi, line 230: Maybe just stick with loading? It's the default for any component/view. If we need separate behaviour (e.g. loadingInclusionPolicy) then those get a different name. All doesn't add much.

frontend/src/components/organizations/OrganizationInfo.vue
```diff
@@ -216,55 +221,95 @@ export default {
...
+      type: Boolean,
+      default: false,
+    },
+    loadingInclusionPolicy: {
```
weskubo-cgi, line 234: If we need to keep this flag, I don't think it needs to be a prop. You can just manage it within the save method directly.

frontend/src/components/organizations/OrganizationInfo.vue
```diff
@@ -216,55 +221,95 @@ export default {
...
+      this.$emit('saveInclusionPolicyData', this.organizationEdit, this.documentsToUpload, this.documentsToDelete)
     },
 
     toggleEditMode() {
-      this.organizationEdit = JSON.parse(JSON.stringify(this.organization))
       this.editMode = !this.editMode
+      this.showUploadDocumentsAlert = false
+      if (!this.editMode) {
+        this.organizationEdit.hasInclusionPolicy = this.organization.hasInclusionPolicy
+      }
```
weskubo-cgi, line 279: Is this required? Is this to ensure that the hasInclusionPolicy flag is kept in sync?

frontend/src/views/account-mgmt/ManageOrganizationView.vue
```diff
@@ -84,7 +94,11 @@ export default {
       panelYourFacilities: [0],
       panelOtherFacilities: [0],
       loading: false,
+      loadingInclusionPolicy: false,
       organization: undefined,
+      uploadedDocuments: [],
+      documentsToDelete: [],
+      documentsToUpload: [],
```
weskubo-cgi, line 101: I don't think these need to be in the data. They're only used within Save. You can just pass them around as parameters.

weskubo-cgi (changes requested):

frontend/src/views/account-mgmt/ManageOrganizationView.vue
```diff
@@ -147,8 +170,58 @@ export default {
...
+
+    async saveInclusionPolicyData(updatedOrganization, documentsToUpload, documentsToDelete) {
+      this.loadingInclusionPolicy = true
+      this.documentsToUpload = documentsToUpload
+      this.documentsToDelete = documentsToDelete
+      try {
+        await this.saveOrganization(updatedOrganization)
+        await this.processDocuments()
+        await this.getInclusionPolicyDocuments()
+        this.$refs.organizationInfo.resetData()
```
weskubo-cgi, line 194: This can be done in the OrganizationInfo.vue save.

frontend/src/views/account-mgmt/ManageOrganizationView.vue
```diff
@@ -147,8 +170,58 @@ export default {
...
+        await OrganizationService.updateOrganization(this.organization.organizationId, updatedOrganization)
+        this.organization = updatedOrganization
+      } catch (error) {
+        this.setFailureAlert('Failed update Inclusion Policy on Organization: ' + this.organization.organizationId, error)
+        return
+      }
     },
+
+    async saveInclusionPolicyData(updatedOrganization, documentsToUpload, documentsToDelete) {
+      this.loadingInclusionPolicy = true
```
weskubo-cgi, line 187: This can be managed in the child.

weskubo-cgi (changes requested):

frontend/src/components/organizations/OrganizationInfo.vue
jgstorey: Fixed.

frontend/src/views/account-mgmt/ManageOrganizationView.vue
jgstorey: Fixed.

frontend/src/views/account-mgmt/ManageOrganizationView.vue
jgstorey: Fixed.

frontend/src/components/organizations/OrganizationInfo.vue
jgstorey: Fixed. re: initalizeData() with the new changes it clearly revealed we need to reset documents. Not sure why was working before. Regardless fixed.

frontend/src/components/organizations/OrganizationInfo.vue
jgstorey: Replaced with initalizeData() as discussed.

frontend/src/views/account-mgmt/ManageOrganizationView.vue
jgstorey: Fixed.

weskubo-cgi (approved):

PR #120, opened 2024-03-06
Ofmcc 1624 support needs

jenbeckett:


frontend/src/views/supp-allowances/SupplementaryFormView.vue
jgstorey, line 1: <v-col cols="12" align="right">

'align' isn't a valid prop for v-col, use class="text-right" instead.

jgstorey (changes requested):

frontend/src/views/supp-allowances/SupplementaryFormView.vue
jenbeckett: done :)

jgstorey (approved):
Reviewed and approved.

PR #123, opened 2024-03-09
ofmcc-1780 - change update event to make AppDocumentUpload work with …

vietle-cgi:
change update event to make AppDocumentUpload work with v-model


jgstorey (approved):
Confirming reviewed.

PR #128, opened 2024-03-13
Ofmcc 1839 - Update Application/Licence declaration

vietle-cgi:
CRM team updated the field type of Licence/Application declaration, so we need to make the same update. https://eccbc.atlassian.net/browse/OFMCC-1655


jgstorey (approved):
Confirmed reviewed.

PR #132, opened 2024-03-15
bugfix so apps will save again

jenbeckett:


jgstorey (approved):
Reviewed approved.

frontend/src/services/applicationService.js
```diff
@@ -98,6 +98,32 @@ export default {
     }
   },
 
+  async getSupplementaryApplicationsForForm(applicationId) {
+    try {
+      if (!applicationId) return
+      const url = `${ApiRoutes.SUPPLEMENTARY_APPLICATIONS}/${applicationId}`
+
+      const response = await ApiService.apiAxios.get(url)
+      //clean up the fields I don't need
```
weskubo-cgi, line 107: Would it make more sense to just include the fields you do need? What if I new field is added? Will it cause issues if not deleted here?

weskubo-cgi (changes requested):

frontend/src/services/applicationService.js
jenbeckett: nope -it was leftover from the other code - it's gone now

weskubo-cgi: Discussed with Jen and there are more fields retained than deleted so the current approach makes sense.

weskubo-cgi (approved):

PR #136, opened 2024-03-18
Default application type to 'OFM' in CancelApplicationDialog.vue

jgstorey:


frontend/src/components/applications/CancelApplicationDialog.vue
```diff
@@ -60,6 +60,9 @@ export default {
       },
     },
   },
+  created() {
```
weskubo-cgi, line 63: Is there any reason we have a constant and a hard-coded default in applicationType? If it's only used in applicationType there's no need for a constant. If it's used elsewhere then applicationType should use the constant.

weskubo-cgi (changes requested):

frontend/src/components/applications/CancelApplicationDialog.vue
jgstorey: Meant to get rid of the constant. Fixed.

weskubo-cgi (approved):

PR #144, opened 2024-03-27
Fixed hard coded sorting of request sub types... now using ofm_displa…

jgstorey:
…y_order from the backend.

### DIFF
--- a/backend/src/components/lookup.js
+++ b/backend/src/components/lookup.js
@@ -25,7 +25,7 @@ async function getRequestSubCategories() {
   let requestSubCategories = lookupCache.get('requestSubCategories')
   if (!requestSubCategories) {
     requestSubCategories = []
-    let response = await getOperation('ofm_subcategories')
+    let response = await getOperation('ofm_subcategories?$select=ofm_subcategoryid,ofm_name&$orderby=ofm_display_order asc')
     response?.value?.forEach((item) => requestSubCategories.push(new MappableObjectForFront(item, RequestSubCategoryMappings)))
     lookupCache.put('requestSubCategories', requestSubCategories, ONE_HOUR_MS)
   }
--- a/backend/src/components/lookup.js
+++ b/backend/src/components/lookup.js
@@ -14,7 +14,7 @@ async function getRequestCategories() {
   let requestCategories = lookupCache.get('requestCategories')
   if (!requestCategories) {
     requestCategories = []
-    let response = await getOperation('ofm_request_categories')
+    const response = await getOperation('ofm_request_categories')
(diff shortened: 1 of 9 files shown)


frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -395,15 +396,7 @@ export default {
...
-    this.requestSubCategorySortOrder = [
-      "Organization Details",
-      "Facility Details",
-      "Organization phone/email",
-      "Facility phone/email",
-      "Add/change a licence",
-      "Other"
-    ]
-    this.sortedRequestSubCategories = this.sortRequestSubCategories(this.requestSubCategories, this.requestSubCategorySortOrder)
+    this.sortedRequestSubCategories = this.requestSubCategories.sort((a, b) => a.displayOrder - b.displayOrder)
```
vietle-cgi, line 399: instead of sort subcategories in frontend, can we add this one to backend? also please use const instead of let.
const response = await getOperation('ofm_subcategories?$orderby=ofm_display_order')

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -346,6 +346,7 @@ export default {
       return this.isSubCategoryChecked(REQUEST_SUB_CATEGORY_NAMES.ORGANIZATION_PHONE_EMAIL)
     },
     isFacilityPhoneEmailChecked() {
+      console.log(this.isSubCategoryChecked(REQUEST_SUB_CATEGORY_NAMES.FACILITY_PHONE_EMAIL))
```
vietle-cgi, line 349: can we remove this log?

vietle-cgi (changes requested):

frontend/src/components/messages/NewRequestDialog.vue
jgstorey: Fixed.

frontend/src/components/messages/NewRequestDialog.vue
jgstorey: Doh! Fixed.

backend/src/components/lookup.js
```diff
@@ -25,7 +25,7 @@ async function getRequestSubCategories() {
   let requestSubCategories = lookupCache.get('requestSubCategories')
   if (!requestSubCategories) {
     requestSubCategories = []
-    let response = await getOperation('ofm_subcategories')
+    let response = await getOperation('ofm_subcategories?$orderby=ofm_display_order')
```
vietle-cgi, line 28: can we change this response from let to const?

vietle-cgi (approved):

PR #154, opened 2024-04-02
unique vin complete

jenbeckett:


frontend/src/utils/common.js
```diff
@@ -18,3 +18,18 @@ export function isApplicationLocked(applicationStatusCode) {
   }
   return !(applicationStatusCode === SUPPLEMENTARY_APPLICATION_STATUS_CODES.DRAFT || applicationStatusCode === SUPPLEMENTARY_APPLICATION_STATUS_CODES.ACTION_REQUIRED)
 }
+
+export function hasDuplicateVIN(model, transportModels) {
```
weskubo-cgi, line 22: Can we add a quick comment explaining what this method does? It's not immediately obvious that we are checking > 1 because you'll always get a match to yourself.
Code can also be simplified using filter: `transportModels.filter(m => m.VIN === model.VIN).length > 1`

weskubo-cgi (changes requested):

frontend/src/utils/common.js
jenbeckett: done!

weskubo-cgi (approved):

PR #166, opened 2024-04-16
Permission framework

weskubo-cgi:
Initial implementation of permission based security framework. Includes the following:

- Securing a route by permission
- Securing a component (e.g button) by permission
- Using permissions to make a screen read only
- Securing an endpoint by permission by storing the role in the jwt

Known issues:

- There's sometimes a timing issue where the roles (in lookupData) are loaded after the hasPermission check is run. This seems to occur mostly on initial startup of the backend when the values aren't cached. Open to workarounds otherwise will monitor for severity. The query should be much faster in OpenShift so the issue may not appear.

The following is still pending:

- Impersonate roles
- Full security matrix implementation

### DIFF
--- a/backend/src/app.js
+++ b/backend/src/app.js
@@ -61,7 +61,7 @@ app.use(
 
 const logStream = {
   write: (message) => {
-    log.info(message)
+    log.http(message)
   },
 }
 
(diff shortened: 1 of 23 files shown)


weskubo-cgi:
All changes are in https://github.com/bcgov/ecc-ofm/pull/169.

PR #184, opened 2024-04-30
Add ConfigMaps for portal backend

trev-dev:
Jira: OFMCC-1860


tools/config/update-configmap.sh
```diff
@@ -0,0 +1,88 @@
...
+echo Formatting public key from SOAM
+formattedPublicKey=$(cat << PUBKEY
+-----BEGIN PUBLIC KEY-----
+$(echo "$soamOneLineKey" | sed -e "s/.\{64\}/&\n/g")
+-----END PUBLIC KEY-----
+PUBKEY
+)
+echo "$formattedPublicKey"
+
+getSecret(){
```
weskubo-cgi, line 45: Is this used?

tools/config/update-configmap.sh
```diff
@@ -0,0 +1,88 @@
...
+
+echo Removing key files
+rm tempPenBackendkey
+rm tempPenBackendkey.pub
+
+echo Creating config map "$APP_NAME-backend-config-map"
+oc create -n "$OPENSHIFT_NAMESPACE-$envValue" configmap \
+  "$APP_NAME-backend-$envValue-config-map" \
+  --from-literal="CLAMAV_HOST=clamav.$COMMON_NAMESPACE-$envValue.svc.cluster.local" \
+  --from-literal=CLAMAV_PORT=3310 \
```
weskubo-cgi, line 62: We should be consistent about our usage of quotes. See others below as well.

weskubo-cgi (changes requested):

tools/config/update-configmap.sh
trev-dev: > We should be consistent about our usage of quotes. See others below as well.

The quotation actually came from my tooling as a good suggestion, as quoting prevents accidental glob/shell expansions

tools/config/update-configmap.sh
```diff
@@ -0,0 +1,108 @@
+set -euo pipefail
+
+envValue=$1
```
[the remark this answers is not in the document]
trev-dev: @weskubo-cgi copy-pasta.  I agree, let's upcase this

tools/config/update-configmap.sh
trev-dev: Nope

tools/config/update-configmap.sh
weskubo-cgi: I got it. Didn't notice that the unquoted entries have no vars. Thanks!

weskubo-cgi (approved):

weskubo-cgi (approved):

weskubo-cgi:
Latest updates look good. Dev and Test (QA) environments have been validated.

PR #192, opened 2024-05-09
Ofmcc 2037 funding agreement view

jgstorey:
*** NOTE ***
1) Date validation when using 'Custom' date filter option is still subject to change and there is a console error. Need to confirm expected validation behaviors with Danna before i go further, this will be done before PR approved. Didn't want to hold up initial 1st pass of code QA.
2) Things i could spent more time on but need to get sense of value in doing so:
a) Added component AppButtonRadioGroup.vue for radio button options. This is very similar but more generic than AppYesNoInput component. We could remove and refactor with AppButtonRadioGroup though will require some tweaks.
b) Could make content of v-window-item tab value 'agreements' a component.


backend/src/util/common.js
```diff
@@ -12,6 +13,21 @@ function buildFilterQuery(query, mapping) {
   return filterQuery
 }
 
+// Dynamics 365 expects OData queries on dates to be in ISO format
+function formatToISODateFormat(dateString) {
+  if (moment(dateString, 'YYYY-MM-DD', true).isValid()) {
+    return dateString
+  } else {
```
weskubo-cgi, line 20: You don't need an else block if you're returning from the if.

backend/src/util/common.js
```diff
@@ -12,6 +13,21 @@ function buildFilterQuery(query, mapping) {
...
+    const date = moment(dateString)
+    if (date.isValid()) {
```
weskubo-cgi, line 22: You could do this cleanly with a ternary expression.

frontend/src/services/fundingAgreementService.js
```diff
@@ -49,4 +49,28 @@ export default {
       throw error
     }
   },
+
+  async getFAByFacilityIdAndStartDateThreshold(facilityId, startDateThreshold) {
+    try {
+      if (!facilityId && !startDateThreshold) return
+      let url = `${ApiRoutes.FUNDING_AGREEMENTS}?facilityId=${facilityId}&stateCode=0&startDateThreshold=${startDateThreshold}`
```
weskubo-cgi, line 56: This can be const.

frontend/src/services/fundingAgreementService.js
```diff
@@ -49,4 +49,28 @@ export default {
...
+    } catch (error) {
+      console.log(`Failed to get the list of active funding agreements by facility id and start date threshold - ${error}`)
+      throw error
+    }
+  },
+
+  async getFAByFacilityIdAndStartFromEndDates(facilityId, startDateFrom, startDateTo) {
+    try {
+      if (!facilityId && !(startDateFrom || startDateTo)) return
+      let url = `${ApiRoutes.FUNDING_AGREEMENTS}?facilityId=${facilityId}&stateCode=0&startDateFrom=${startDateFrom}&startDateTo=${startDateTo}`
```
weskubo-cgi, line 68: This can be const.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+      dateFilterTypes: [
+        { label: '3 Months', value: '3Months' },
+        { label: '6 Months', value: '6Months' },
+        { label: 'YTD', value: 'YTD' },
+        { label: 'Custom', value: 'Custom' },
+      ],
+      paymentFilterTypes: [
+        { label: 'Base Funding', value: 'baseFunding' },
+        { label: 'Supplementary Allowances', value: 'supplementaryAllowances' },
+        { label: 'Other', value: 'other' },
```
weskubo-cgi, line 217: These values are repeated on line 240 so they should be constants. If it's just screen specific and not used in the template you can define it in the script but outside of the component.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
         { title: 'Start Date', key: 'startDate' },
         { title: 'End Date', key: 'endDate' },
         { title: 'Status', key: 'status' },
         { title: 'Actions', key: 'actions' },
       ],
+      dateFilterTypes: [
+        { label: '3 Months', value: '3Months' },
+        { label: '6 Months', value: '6Months' },
+        { label: 'YTD', value: 'YTD' },
+        { label: 'Custom', value: 'Custom' },
```
weskubo-cgi, line 212: These values are repeated on line 239 (and elsewhere) so they should be constants. If it's just screen specific and not used in the template you can define it in the script but outside of the component.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+    ...mapState(useAuthStore, ['userInfo']),
+    formattedStartDateFrom() {
+      return this.startDateFrom ? format.formatDate(this.startDateFrom) : ''
+    },
+    formattedStartDateTo() {
+      return this.startDateTo ? format.formatDate(this.startDateTo) : ''
     }
   },
+  async created() {
+    this.FUNDING_AGREEMENT_STATUS_CODES = FUNDING_AGREEMENT_STATUS_CODES
```
weskubo-cgi, line 231: You don't need to define this in created() unless it's required in the template.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+    this.initialize()
+    await this.getOrgUsers() // Retrieve the list of users to retrieve the expense authority name
+    await this.getFundingAgreements()
+  },
+  methods: {
+    initialize() {
```
weskubo-cgi, line 237: Can we use a more specific name like initializeFilter() or the more semantically correct resetFilter()?

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+      this.selectedFacility = null
+      this.selectedDateFilterType = '3Months'
+      this.selectedPaymentFilterTypes = ['baseFunding', 'supplementaryAllowances', 'other']
+      this.startDateFrom = null
+      this.startDateTo = null
+    },
+
+    async getOrgUsers() {
```
weskubo-cgi, line 245: I prefer "load" for methods that aren't just getters or http GET calls. Same applies to `getFundingAgreements()`.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+      try {
+        this.loading = true
+        const res = await ApiService.apiAxios.get(ApiRoutes.USER_PERMISSIONS_FACILITIES + '/' + this.userInfo.organizationId)
```
weskubo-cgi, line 248: Can we create a service method for this? I believe I commented on this elsewhere in the codebase with a TODO so please refactor there as well.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+
+    /** 
+     * TODO: Payment type is displayed in UI but is not yet integrated into this method as required CRM data is not yet in place.
+     * This is a know issue and postponed to a future sprint.
+     */
+    async getFundingAgreements() {
+      try {
+        this.loading = true
+        this.fundingAgreements = []
+        const startDateThreshold = this.startDateThreshold()
```
weskubo-cgi, line 265: There's no benefit to defining a variable here as it's only used once and computed() methods should just be treated as properties.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+        console.error('Error getting expense authority name:', error)
+        return ''
+      }
+    },
+
+    updateSelectedDateFilterType(value) {
+      this.selectedDateFilterType = value
+    },
+
+    startDateThreshold() {
```
weskubo-cgi, line 312: This can be under computed.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+        if (!application?.expenseAuthorityId) return ''
+        const expenseAuthorityUser = this.orgUsers.find(user => user.contactId === application.expenseAuthorityId)
+        return `${expenseAuthorityUser?.firstName} ${expenseAuthorityUser?.lastName}`
+      } catch (error) {
+        console.error('Error getting expense authority name:', error)
+        return ''
+      }
+    },
+
+    updateSelectedDateFilterType(value) {
```
weskubo-cgi, line 308: This is unused.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+        case FUNDING_AGREEMENT_STATUS_CODES.EXPIRED:
+          return 'Expired'
+        default:
+          console.log('Unknown status code:', statusCode)
+          return 'Unknown'
+      }
+    },
+
+    getStatusClass(statusCode) {
+      if (this.FUNDING_AGREEMENT_STATUS_CODES.DRAFT === statusCode) {
```
weskubo-cgi, line 358: You can simplify this code greatly using object binding: https://vuejs.org/guide/essentials/class-and-style.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+        return 'status-gray'
+      } else if ([FUNDING_AGREEMENT_STATUS_CODES.ACTIVE].includes(statusCode)) {
```
weskubo-cgi, line 360: There's no need for an array here.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+        return 'status-green'
+      } else if ([FUNDING_AGREEMENT_STATUS_CODES.SIGNATURE_PENDING].includes(statusCode)) {
```
weskubo-cgi, line 362: There's no need for an array here.

backend/src/components/fundingAgreements.js
```diff
@@ -1,19 +1,33 @@
...
 const HttpStatus = require('http-status-codes')
 const log = require('./logger')
 const { isEmpty } = require('lodash')
 
+function buildFilterQueryDates(queryParams) {
+  if (queryParams?.startDateThreshold) {
+    const startDateThreshold = queryParams.startDateThreshold
+    delete queryParams.startDateThreshold
+    return `ofm_start_date ge ${startDateThreshold} and `
+  } else if (queryParams?.startDateFrom && queryParams?.startDateTo) {
```
weskubo-cgi, line 15: There's no need for an else since the if returns.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+    },
+
+    dateByMonthsInPast(months) {
+      const date = new Date()
+      date.setMonth(date.getMonth() - months)
+      return date.toISOString().split('T')[0]
+    },
+
+    facilityName(facilityId) {
+      return this.userInfo.facilities.find(facility => facility.facilityId === facilityId).facilityName
```
weskubo-cgi, line 332: Is it at all possible there's no match? If so you probably want optional changing before `.facilityName`

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+    },
+
+    statusName(statusCode) {
+      switch (statusCode) {
```
weskubo-cgi, line 336: You could just create a constant in this file mapping status codes to display names and then dynamically lookup the display name by statusCode key.

weskubo-cgi (changes requested):

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -25,21 +25,130 @@
...
+                  {{ format.formatDate(item?.endDate) }}
+                </template>
+                <template #item.status="{ item }">
+                  <span :class="getStatusClass(item?.statusCode)" class="pt-1 pb-1 pl-2 pr-2">{{ item?.status }}</span>
+                </template>
+                <template #[`item.actions`]="{ item }">
+                  <v-btn v-if="item?.status === 'FA Signature Pending'" variant="text" @click="this.$router.push({ name: 'funding', params: { fundingGuid: item.fundingId } })">
+                    <v-icon size="large">mdi-signature-freehand</v-icon>
+                  </v-btn>
+                  <v-btn v-if="['FA Submitted to Ministry', 'Active'].includes(item?.status)" variant="text" @click="this.$router.push({ name: 'funding', params: { fundingGuid: item.guid } })">
```
jenbeckett, line 146: { fundingGuid: item.guid } should be { fundingGuid: item.fundingId } otherwise the link is broken

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +163,211 @@
...
+      } finally {
+        this.loading = false
+      }
+    },
+
+    /** 
+     * TODO: Payment type is displayed in UI but is not yet integrated into this method as required CRM data is not yet in place.
+     * This is a know issue and postponed to a future sprint.
+     */
+    async getFundingAgreements() {
```
jenbeckett, line 261: getFundingAgreements() is always called when we hit the search button. That seems like a lot of calls to dynamics - when we should already have all the funding agreements on load? I feel like this could perhaps be done with a filter on the loaded data so we don't have to hit Dynamics as much.

jenbeckett (changes requested):

backend/src/util/common.js
jgstorey: Fixed.

backend/src/util/common.js
jgstorey: Fixed.

frontend/src/services/fundingAgreementService.js
jgstorey: Fixed.

frontend/src/services/fundingAgreementService.js
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Updated.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: I believe i got this right.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Removed method since redundant now that we are using class binding.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Removed method since redundant now that we are using class binding.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Updated.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: That would def break the link, but branch has been updated to user fundingId since yesterday afternoon when i messaged you the link was working. Let me know if i'm missing something.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: It is a lot of calls to dynamics given we are getting all funding agreements for all user's facilities (when not filtering on facility) and then their related applications. To reduce the calls would have required the planning for ticket to have CRM support. Reducing the calls could further achieved with OData expands added to entities ofm_fundings, _facilities, and _applications with dynamics team support. I would suggest we could flag it to QA for some testing on an org with a high facility count and address further if needed.
cc: @weskubo-cgi

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: You can use a method to return the object so that you don't inline complex logic.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: I wasn't clear enough here. You could keep loadFundingAgreements() and loadOrgUsers() separate if you prefer...just wanted getFoo changed to loadFoo. Given orgUsers is just one line I would keep it code structured as is but rename load() to loadFundingAgreements().

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: If your constants don't need to be access in the template you can just define them outside. See LOCKED_STATUSES in FundingView.vue. It'll keep your code separation cleaner.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +164,233 @@
...
+      return date.toISOString().split('T')[0]
+    },
+
+    facilityName(facilityId) {
+      return this.userInfo.facilities.find(facility => facility.facilityId === facilityId)?.facilityName
+    },
+
+    statusName(statusCode) {
+      const statusName = this.statusNameMap[statusCode];
+      if (statusName) {
```
weskubo-cgi, line 386: I don't think the console.log is necessary. You could just return statusName ?? 'Unknown'.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: Can you update ManageusersView.vue with the new method call?

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: See other comment about constants.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: The issue with in memory filter is that the default is 3 months. If we want to go to a different date range then it will require a call to the server anyways. Unless we preload all data which is probably the worst choice performance wise. I think we leave it as is...I suspect we're going to need to optimize our queries at some point anyways.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -25,21 +25,131 @@
...
-              <template #[`item.actions`]="{ item }">
-                <v-btn v-if="item.status === 'FA Signature Pending'" variant="text" @click="this.$router.push({ name: 'funding', params: { fundingGuid: item.guid } })">
-                  <v-icon size="large">mdi-signature-freehand</v-icon>
-                </v-btn>
-                <v-btn v-if="item.status === 'FA Submitted to Ministry'" variant="text" @click="this.$router.push({ name: 'funding', params: { fundingGuid: item.guid } })">
-                  <v-icon size="large">mdi-folder-open-outline</v-icon>
-                </v-btn>
-              </template>
-            </v-data-table>
+            <v-form ref="faForm" v-model="isFormComplete">
```
weskubo-cgi, line 31: I think the UI needs a little work at lower resolutions:

image

weskubo-cgi (changes requested):

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed. Yikes, i seem to always under estimate this.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: As suggested you can condense the if/else to one line of code `statusName ?? 'Unknown'`. Actually the whole method can be `this.statusNameMap[statusCode] ?? 'Unknown'`.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: Still outstanding.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: Still outstanding.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: Much simpler!

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: Much better but the date range is getting cut off quite early.

image

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Thanks for extra description and reference.... i think i got it now.

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: fixed but has to break quite to single col quite early. Its like there needs to be a sm **md** & lg, not just sm and lg.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +176,236 @@
...
+        return
+      }
+      await this.loadFundingAgreements()
+    },
+
+    async enrichFundingAgreementData(result) {
+      result.fundingAgreementType = this.PAYMENT_FILTER_TYPE_VALUES.BASE_FUNDING
+      result.facility = this.facilityName(result.facilityId)
+      result.status = this.statusName(result.statusCode)
+      result.expenseAuthorityName = await this.getExpenseAuthorityName(result.applicationId)
```
vietle-cgi, line 373: since the FA entity is linked to Application entity, I think we can expand the backend query to include the expenseAuthority instead of having a separate API call to get application. (e.g.: ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode&$expand=ofm_application($select=_ofm_expense_authority_value)&$filter=(_ofm_facility_value eq c9ecef6c-437d-ee11-8179-000d3a09d132 and statecode eq 0)

vietle-cgi (changes requested):

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Fixed.

frontend/src/services/fundingAgreementService.js
```diff
@@ -49,4 +49,28 @@ export default {
...
+      throw error
+    }
+  },
+
+  async getFAByFacilityIdAndStartFromEndDates(facilityId, startDateFrom, startDateTo) {
+    try {
+      if (!facilityId && !(startDateFrom || startDateTo)) return
+      const url = `${ApiRoutes.FUNDING_AGREEMENTS}?facilityId=${facilityId}&stateCode=0&startDateFrom=${startDateFrom}&startDateTo=${startDateTo}`
+      const response = await ApiService.apiAxios.get(url)
+      return response?.data[0]
```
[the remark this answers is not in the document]
jgstorey: Excellent catch! Fixed.

frontend/src/services/fundingAgreementService.js
```diff
@@ -49,4 +49,28 @@ export default {
       throw error
     }
   },
+
+  async getFAByFacilityIdAndStartDateThreshold(facilityId, startDateThreshold) {
+    try {
+      if (!facilityId && !startDateThreshold) return
+      const url = `${ApiRoutes.FUNDING_AGREEMENTS}?facilityId=${facilityId}&stateCode=0&startDateThreshold=${startDateThreshold}`
+      const response = await ApiService.apiAxios.get(url)
+      return response?.data[0]
```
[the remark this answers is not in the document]
jgstorey: Fixed.

frontend/src/views/account-mgmt/ManageUsersView.vue
```diff
@@ -169,7 +168,7 @@ export default {
       try {
         this.loading = true
         // TODO (jgstorey) Make this a service function
```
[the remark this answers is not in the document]
jgstorey: Fixed.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -54,47 +176,236 @@
...
+    /** 
+     * TODO: Payment type is displayed in UI but is not yet integrated into this method as required CRM data is not yet in place.
+     * This is a know issue and postponed to a future sprint.
+     */
+    async loadFundingAgreements() {
+      try {
+        this.loading = true
+        this.fundingAgreements = []
+        const facilities = this.selectedFacility ? [this.selectedFacility] : this.userInfo.facilities
+        for (let facility of facilities) {
```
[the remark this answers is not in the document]
jgstorey: If this doesn't improve performance, i would suggest we leave it as working? We need this in QA and I need to move on some other project work unless some else can pick it up. cc: @weskubo-cgi

frontend/src/views/funding/FundingOverviewView.vue
jgstorey: Correct, the backend calls can be reduced with expands in the OData query, though this will require CRM developer support to add. This is what Wes's team comment that we'll likely need to optimize this in the future. Suggest we leave for now as Wes's suggestion to address later.

frontend/src/views/funding/FundingOverviewView.vue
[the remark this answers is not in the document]
weskubo-cgi: Just add a TODO for now.

frontend/src/views/funding/FundingOverviewView.vue
weskubo-cgi: I'm gonna add this suggestion to the Funding agreement tech spike.

frontend/src/views/funding/FundingOverviewView.vue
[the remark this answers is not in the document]
jgstorey: Added TODO

frontend/src/views/funding/FundingOverviewView.vue
vietle-cgi: when I try this query, I can see the expenseAuthorityId in the response. I think it is sufficient?
ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode&$expand=ofm_application($select=_ofm_expense_authority_value)&$filter=(_ofm_facility_value eq c9ecef6c-437d-ee11-8179-000d3a09d132 and statecode eq 0)

vietle-cgi: image

vietle-cgi (approved):

weskubo-cgi (approved):

jenbeckett (approved):

PR #195, opened 2024-05-13
OFMCC-2048 - add Employee Certificates to Staffing page in Core Application

vietle-cgi:
Summary of this ticket:
- Add Employee Certificate input to Staffing view and Staffing Summary view
- Add some validations for Employee Certificate: The provider must provide ECE/ECEA certificates for all employees (the number of staff and the number of certificates must match, and there must be no duplicate Initials or Certificate Numbers)
- The max length of Initials is 4, and the max length of Certificate Numbers is 6. Certificate Numbers is a number field

### DIFF
--- a/frontend/src/assets/css/main.css
+++ b/frontend/src/assets/css/main.css
@@ -10,74 +10,12 @@ input[type='number'] {
   -moz-appearance: textfield;
 }
 
-.noticeAlert {
-  font-size: medium;
-  color: #d8292f;
-  font-family: 'BCSans', Verdana, Arial, sans-serif;
-  padding-top: 8px;
-  padding-bottom: 8px;
-  background-color: #f2dede;
-  border: 1px solid #d8292f;
-}
-
-.noticeWaringIcon {
-  color: #6c4a00 !important;
-}
-.noticeWarning {
-  font-size: medium;
-  color: #6c4a00;
-  font-family: 'BCSans', Verdana, Arial, sans-serif;
-  padding-top: 8px;
-  padding-bottom: 8px;
-  background-color: #f9f1c6;
-  border: 1px solid #6c4a00;
-}
-
-.noticeInfoIcon {
-  color: #003366 !important;
-}
-.noticeInfo {
-  font-size: medium;
-  color: #313132;
-  font-family: 'BCSans', Verdana, Arial, sans-serif;
-  padding-top: 8px;
-  padding-bottom: 8px;
-  background-color: #c1dcf6;
-  border: 1px solid #313132;
-}
-
-.disabledButton {
-  background-color: rgba(0, 0, 0, 0.12) !important;
-  color: rgba(0, 0, 0, 0.26) !important;
-  box-shadow: none !important;
-}
-
-.blueButton {
-  color: white !important;
-  background-color: #003366 !important;
-}
-
-.stickyNavButtons {
-  position: sticky;
-  bottom: 0;
-  padding: 10px 0px;
-  background-color: lightgray;
-  z-index: 2;
-}
-
 @keyframes bone {
   to {
     background-position: 100% 0;
   }
 }
 
-#toTopBtn {
-  opacity: 0.5;
-}
-
-#toTopBtn:hover {
-  opacity: 1;
-}
 .v-alert.bootstrap-success {
   color: #234720 !important;
   background-color: #d9e7d8 !important;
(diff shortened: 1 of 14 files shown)


frontend/src/components/applications/StaffingCertificateInput.vue
```diff
@@ -0,0 +1,198 @@
...
+              @input="update(certificate)"
+              @blur="checkDuplicateInitials" />
+          </v-col>
+          <v-col cols="12" sm="6" class="pt-1">
+            <v-text-field
+              v-model="certificate.certificateNumber"
+              variant="outlined"
+              density="compact"
+              :disabled="readonly"
+              :placeholder="certificateNumberPlaceholder"
```
weskubo-cgi, line 27: The placeholder is being cut off:

image

frontend/src/assets/css/main.css
```diff
@@ -10,74 +10,12 @@ input[type='number'] {
   -moz-appearance: textfield;
 }
 
-.noticeAlert {
```
weskubo-cgi, line 13: Thanks for doing some cleanup!

frontend/src/components/applications/StaffingCertificateInput.vue
```diff
@@ -0,0 +1,198 @@
...
+              maxlength="6"
+              rounded="0"
+              :rules="rules.required"
+              :error-messages="getErrorMessagesForDuplicateError(certificate.isCertificateNumberDuplicate)"
+              @input="update(certificate)"
+              @blur="checkDuplicateCertificateNumbers" />
+          </v-col>
+          <v-col v-if="certificate.initials || certificate.certificateNumber" cols="12" sm="1" class="pl-3" align="center">
+            <v-btn icon="mdi-trash-can-outline" variant="text" @click="remove(index)" />
```
weskubo-cgi, line 36: The v-btn has a weird squished oval overlay on hover:
image. Maybe try with child icon which looks on on ApplicationsHistoryView.vue.

frontend/src/services/applicationService.js
```diff
@@ -183,4 +183,41 @@ export default {
     )
     return results.some((result) => result)
   },
+
+  async createEmployeeCertificate(payload) {
+    try {
+      if (isEmpty(payload)) return
+      const response = await ApiService.apiAxios.post(`${ApiRoutes.APPLICATIONS}/employeeCertificate`, payload)
```
weskubo-cgi, line 190: Can we create service methods for these instead of calling the generic ApiService?

frontend/src/components/applications/StaffingCertificateInput.vue
```diff
@@ -0,0 +1,198 @@
...
+      type: Number,
+      required: true,
+    },
+    allCertificates: {
+      type: Array,
+      default: () => [],
+    },
+  },
+
+  emits: ['update:modelValue'],
```
weskubo-cgi, line 73: I think the use of v-model with an array is problematic as we see in the console.

weskubo-cgi (changes requested):

frontend/src/services/applicationService.js
vietle-cgi: sorry @weskubo-cgi, probably my brain is not working on Monday. Could you please clarify this comment? The createEmployeeCertificate() is a service method in applicationService.

weskubo-cgi: No, I'm losing it. Didn't notice what file I was looking at!

frontend/src/components/applications/StaffingCertificateInput.vue
vietle-cgi: yes, the placeholder is being cut off. If we want to always display the full placeholder text, we will need to reduce the number of columns, which will make the page really long when the provider has many staff (e.g.: 99 staffs --> 99 certificates). I will reduce the number of columns. If the clients complain about the length of the page, we can do an enhancement ticket to shorten the placeholder text or add a scrollbar to the certificates list.

frontend/src/components/applications/StaffingCertificateInput.vue
vietle-cgi: it's updated

frontend/src/components/applications/StaffingCertificateInput.vue
vietle-cgi: it's updated to use emit event instead of v-model. There is no more warning message. Thanks for your suggestions!!

frontend/src/components/applications/StaffingCertificateInput.vue
vietle-cgi: I've updated the placeholder text to "ECE certificate #"

weskubo-cgi (approved):

PR #201, opened 2024-05-18
ofmcc-3437 - prevent users from creating a new application if there is an active application

vietle-cgi:
Prevent users from creating a new application if there is an active application:
In Select Facility Page, when the user clicks Next, we will check if there are any active applications:
--> Yes: check if there is any DRAFT application 
                   --> Yes: redirect user to DRAFT application
                   --> No: redirect user to the latest Active application (using latestActivityDate)
--> No: create a new application

### DIFF
--- a/frontend/src/components/applications/StaffingCertificateInput.vue
+++ b/frontend/src/components/applications/StaffingCertificateInput.vue
@@ -32,7 +32,7 @@
               @input="update(certificate)"
               @blur="checkDuplicateCertificateNumbers" />
           </v-col>
-          <v-col v-if="certificate.initials || certificate.certificateNumber" cols="12" sm="1">
+          <v-col v-if="!readonly && (certificate.initials || certificate.certificateNumber)" cols="12" sm="1">
             <v-btn variant="text" @click="remove(index)">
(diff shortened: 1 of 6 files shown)


frontend/src/components/applications/StaffingCertificateInput.vue
```diff
@@ -32,7 +32,7 @@
               @input="update(certificate)"
               @blur="checkDuplicateCertificateNumbers" />
           </v-col>
-          <v-col v-if="certificate.initials || certificate.certificateNumber" cols="12" sm="1">
+          <v-col v-if="!readonly && (certificate.initials || certificate.certificateNumber)" cols="12" sm="1">
```
weskubo-cgi, line 35: Given that there's three conditions here I'd be inclined to move this to a method. E.g. showTrash().

weskubo-cgi (changes requested):

frontend/src/components/applications/StaffingCertificateInput.vue
vietle-cgi: it's updated.

weskubo-cgi (approved):

PR #206, opened 2024-05-23
OFMCC-2053 - Add CCOF enrolment check for adding core application

vietle-cgi:
Add CCOF enrolment check for adding core application

### DIFF
--- a/frontend/src/stores/auth.js
+++ b/frontend/src/stores/auth.js
@@ -3,8 +3,34 @@ import { defineStore } from 'pinia'
 
 import ApiService from '@/common/apiService'
 import AuthService from '@/common/authService'
+import FacilityService from '@/services/facilityService'
 import { useAppStore } from '@/stores/app'
-import { APPLICATION_INTAKE_TYPES } from '@/utils/constants'
+import { APPLICATION_INTAKE_TYPES, OFM_PROGRAM_CODES } from '@/utils/constants'
+
+/*
+  A facility can apply for a core application if it satisfies these conditions:
+  1. There is an Open Intake or the facility is in the allowed facility list of a Limited Intake
+  2. If the facility is in the CCOF program, their program start date must >= 1 year from the current date.
+*/
+async function checkAddCoreApplicationEligibility(facility) {
+  try {
+    const appStore = useAppStore()
+    const facilityInfo = await FacilityService.getFacility(facility?.facilityId)
+    // TODO (vietle-cgi) - include TDAD program to programCheck
+    const programCheck = facilityInfo?.programCode === OFM_PROGRAM_CODES.OFM || (facilityInfo?.programCode === OFM_PROGRAM_CODES.CCOF && facilityInfo?.ccofOneYearEnrolment)
+
+    const intakeWindowCheck = appStore.applicationIntakes?.some((intake) => {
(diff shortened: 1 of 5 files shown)


frontend/src/views/applications/ApplicationsHistoryView.vue
```diff
@@ -21,7 +21,10 @@
             </v-card-title>
             <v-card-text class="text-center d-flex flex-column align-center pt-4 pb-0">
               <div v-if="isAddCoreApplicationAllowed">Before starting an application, verify your organization and facility details in Account Management.</div>
-              <div v-else>The Application intake is currently closed.</div>
+              <div v-else>
+                <div v-if="!isCCOFEnrolmentCheckSatisfied">Enrolment in CCOF for greater than 1 year is required to apply.</div>
```
weskubo-cgi, line 25: I might be tempted to use either a ternary expression with {{}} or a computed method to determine the text. It will avoid the extra level of nesting and unnecessary divs.

frontend/src/stores/auth.js
```diff
@@ -3,8 +3,36 @@ import { defineStore } from 'pinia'
...
 import { useAppStore } from '@/stores/app'
-import { APPLICATION_INTAKE_TYPES } from '@/utils/constants'
+import { APPLICATION_INTAKE_TYPES, OFM_PROGRAM_CODES } from '@/utils/constants'
+
+/*
+  A facility can apply for a core application if it satisfies these conditions:
+  1. There is an Open Intake or the facility is in the allowed facility list of a Limited Intake
+  2. If the facility is in the CCOF program, their program start date must >= 1 year from the current date.
+*/
+function intakeWindowCheckForAddApplication(facility) {
```
weskubo-cgi, line 15: This is being picky but can we put these utility functions at the bottom of the file? It was confusing at first as to why they were in the store until I realized they were used in getUserInfo().

frontend/src/stores/auth.js
```diff
@@ -3,8 +3,36 @@ import { defineStore } from 'pinia'
...
+    const isOpenIntake = intake.type === APPLICATION_INTAKE_TYPES.OPEN_INTAKE
+    const limitedIntakeFacilities = intake?.facilities?.map((facility) => facility.facilityId)
+    return isWithinApplicationIntakeWindow && (isOpenIntake || limitedIntakeFacilities?.includes(facility.facilityId))
+  })
+  return intakeWindowCheck
+}
+
+async function ccofEnrolmentCheckForAddApplication(facility) {
+  try {
+    const facilityInfo = await FacilityService.getFacility(facility?.facilityId)
```
weskubo-cgi, line 28: Would it make more sense to have ofm_ccof_requirement/ccofOneYearEnrolment available as part of the ProviderProfile payload? It'll save a query for each facility.

weskubo-cgi (changes requested):

frontend/src/stores/auth.js
vietle-cgi: it's updated. thanks for your suggestions!

frontend/src/stores/auth.js
vietle-cgi: it's updated. thanks for your suggestions!

weskubo-cgi (approved):

PR #217, opened 2024-05-29
OFMCC-1830 - Funding permissions

weskubo-cgi:
- Removed Documents from Home Page
- Added ability to hide routes
- Added funding security
- Added missing licenses security
- Updated validatePermission to accept a list of permissions
- Added backend validation for signing funding agreement

### DIFF
--- a/backend/src/middlewares/validateExpenseAuthority.js
+++ b/backend/src/middlewares/validateExpenseAuthority.js
@@ -12,13 +12,12 @@ module.exports = function () {
 
     const contactId = req?.body.contactId
     // Only validate a signing update (which will include contactId)
-    if (!contactId) next()
+    if (!contactId) return next()
 
     const user = req.session?.passport?.user
     // Validate that the contactId represents the currentUser as you can only sign for yourself
     if (contactId !== user?.contactId) {
-      res.sendStatus(403)
-      return
+      return res.sendStatus(403)
     }
 
     // Send raw queries since we don't want to mess with the request/response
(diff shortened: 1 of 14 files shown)


backend/src/middlewares/validateExpenseAuthority.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+module.exports = function () {
+  return async function (req, res, next) {
+    log.verbose(`validating expense authority`)
+
+    const contactId = req.session?.passport?.user?.contactId 
+    // Validate that the contactId represents the currentUser as you can only sign for yourself
+    if (req?.body.contactId !== contactId) {
+      res.sendStatus(403)
+      return
```
trev-dev, line 21: Do we mean to explicitly return `undefined` here?  Why not `return res.sendStatus(403)`?

backend/src/middlewares/validateExpenseAuthority.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param
+ * @returns
+ */
+module.exports = function () {
+  return async function (req, res, next) {
+    log.verbose(`validating expense authority`)
+
+    const contactId = req?.body.contactId
+    // Only validate a signing update (which will include contactId)
+    if (!contactId) next()
```
trev-dev, line 15: If there's no contactId, do we maybe want to return `next()` here to stop the flow early?

weskubo-cgi: Yes. This was missed.

trev-dev (approved):

PR #218, opened 2024-05-29
Ofmcc 3821 multiple addresses

jenbeckett:


backend/src/components/facilities.js
```diff
@@ -1,14 +1,56 @@
...
-const { FacilityMappings, RoleMappings, UserMappings, UsersPermissionsFacilityMappings, LicenceMappings } = require('../util/mapping/Mappings')
+const { FacilityMappings, RoleMappings, UserMappings, UsersPermissionsFacilityMappings, LicenceMappings, FacilityAdditionalAddressMappings } = require('../util/mapping/Mappings')
 const HttpStatus = require('http-status-codes')
 
+/**
+ * Dynamics can take a selector with a list of fields
+ * Use this function to generate a list of fields based on mappings
+ */
+function getMappingString(mappings) {
+  let retVal = mappings
```
weskubo-cgi, line 12: You can write this as a pretty clear single line if you skip the variable declaration and use an implicit return.
This might be useful elsewhere in the code...can we put it somewhere common?

backend/src/components/facilities.js
```diff
@@ -1,14 +1,56 @@
...
+    .map((item) => {
+      return item.back
+    })
+    .join(',')
+  return retVal
+}
+
+function makeThePayloadPretty(response) {
```
weskubo-cgi, line 20: How about just formatPayload or mapPayload/convertPayload?

backend/src/util/mapping/Mappings.js
```diff
@@ -243,7 +244,74 @@ const FacilityMappings = [
   { back: 'statecode', front: 'stateCode' },
   { back: 'statuscode', front: 'statusCode' },
   { back: '_ofm_primarycontact_value', front: 'primaryContactId' },
-  { back: 'ofm_program', front: 'programCode' },
+]
+const FacilityAdditionalAddressMappings = [
```
weskubo-cgi, line 248: I'm questioning whether it's worth the benefit of storing this as a Mapping if we're not using it to format for front. We could just generate the query string using a for loop, like how you handle the response.

weskubo-cgi (changes requested):

frontend/src/components/facilities/FacilityInfo.vue
```diff
@@ -170,6 +170,78 @@
             </v-row>
           </v-col>
         </v-row>
+
+        <v-divider />
```
vietle-cgi, line 174: if a facility has no additional addresses, we'll have a redundant divider. can we put both divider and expansion-panels inside a div then add v-if to that div? Or can we remove that v-divider because it doesn't look much different if we have it or not?

backend/src/components/facilities.js
jenbeckett: whoops I meant to change this before I pushed the code !! done

frontend/src/components/facilities/FacilityInfo.vue
```diff
@@ -170,6 +170,78 @@
...
+
+        <v-expansion-panels v-if="facility?.additionalAddresses.length > 0">
```
vietle-cgi, line 176: since this expansion panel is for physical address, I think we should move it to right under physical address. Right now it looks like we have multiple mailing addresses. 
image

vietle-cgi (changes requested):

backend/src/components/facilities.js
jenbeckett: done! Yes I thought you might want to use it in other places. Could be worth refactoring some of the giant strings to use this instead at some point

frontend/src/components/facilities/FacilityInfo.vue
jenbeckett: you are right good point - i'll fix this

frontend/src/components/facilities/FacilityInfo.vue
jenbeckett: yup and done!

frontend/src/components/facilities/FacilityInfo.vue
```diff
@@ -109,6 +109,77 @@
           </v-col>
         </v-row>
         <v-divider />
```
vietle-cgi, line 111: should we move this v-divider to under the expansion-panel?

vietle-cgi (changes requested):

frontend/src/components/facilities/FacilityInfo.vue
vietle-cgi: and maybe add some padding top/bottom so it is more visible

backend/src/components/facilities.js
weskubo-cgi: You can simplify this further with an implicit return:

`return mappings.map((item) => item.back).join(',')`

frontend/src/components/facilities/FacilityInfo.vue
jenbeckett: I'm just going to hide the divider if we have the expansion panel... I don't feel like we really need it in that case

backend/src/components/facilities.js
jenbeckett: yup done now

backend/src/util/mapping/Mappings.js
jenbeckett: it's been moved out!

vietle-cgi (approved):

weskubo-cgi (approved):

PR #228, opened 2024-06-06
OFMCC-4394 - Deactivate user not removing role

weskubo-cgi:
Fixed role handling by updating request structure.

### DIFF
--- a/backend/src/components/user.js
+++ b/backend/src/components/user.js
@@ -264,8 +264,15 @@ function mapContactForBack(data) {
   const contact = new MappableObjectForBack(data, ContactMappings).toJSON()
   contact.entityNameSet = 'contacts'
   contact.actionMode = 'Update'
-  contact['ofm_portal_role_id@odata.bind'] = `/ofm_portal_roles(${data?.role?.roleId})`
+
+  // Handle role Update/Delete
+  if (contact.role) {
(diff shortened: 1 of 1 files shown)


backend/src/components/user.js
```diff
@@ -264,8 +264,15 @@ function mapContactForBack(data) {
   const contact = new MappableObjectForBack(data, ContactMappings).toJSON()
   contact.entityNameSet = 'contacts'
   contact.actionMode = 'Update'
-  contact['ofm_portal_role_id@odata.bind'] = `/ofm_portal_roles(${data?.role?.roleId})`
+
+  // Handle role Update/Delete
+  if (contact.role) {
+    contact['ofm_portal_role_id@odata.bind'] = `/ofm_portal_roles(${contact.role.roleId})`
+  } else {
```
vietle-cgi, line 271: Is this else statement necessary? should this if else statement be like:
```
if (contact.role) {
    contact['ofm_portal_role_id@odata.bind'] = `/ofm_portal_roles(${contact.role.roleId})`
    delete contact.role
  }
```

vietle-cgi (changes requested):

backend/src/components/user.js
weskubo-cgi: There's different syntax to remove the role.

vietle-cgi (approved):

PR #229, opened 2024-06-06
OFMCC-3823 - View Monthly Reporting table

vietle-cgi:
Changes in this PR:
1. Fix date format warning in buildDateFilterQuery in backend
2. Refactor Monthly reporting:
- Remove Create Report Response from Portal: every month, there is a flow in CRM triggered to create Draft Report response
- Remove report question types, months, fiscal years from lookup
- Update business rule field in Get Report Questions query in backend to match with CRM change
- Remove report types and question occurrence because we only have Monthly Reporting for MVP. We will revisit it once we have new report templates.


frontend/src/components/reports/CancelSurveyResponseDialog.vue
```diff
@@ -1,15 +1,15 @@
 <template>
-  <AppDialog v-model="isDisplayed" title="Confirm" :isLoading="isLoading" persistent max-width="50%" @close="closeDialog">
+  <AppDialog v-model="isDisplayed" title="Confirm" :isLoading="loading" persistent max-width="50%" @close="closeDialog">
     <template #content>
       <div class="confirm-dialog-text d-flex flex-column align-center">Are you sure you want to cancel this report response?</div>
```
weskubo-cgi, line 4: This might be handled with the Delete story but we might want to update the text here. Not sure report response makes sense to the user.

frontend/src/components/reports/PendingReportsTab.vue
```diff
@@ -5,73 +5,75 @@
...
+        <template #[`item.title`]="{ item }">
+          <span>{{ getReportTitle(item) }}</span>
+        </template>
+        <template #[`item.status`]>
+          <span class="status-gray">Draft</span>
         </template>
         <template #[`item.actions`]="{ item }">
-          <v-btn v-if="showOpen(item)" icon="mdi-folder-open-outline" variant="text" @click="openSurveyResponse(item)" />
-          <v-btn v-if="showTrash(item)" icon="mdi-trash-can-outline" variant="text" @click="toggleCancelDialog" />
+          <v-btn variant="text" @click="openSurveyResponse(item)">
```
weskubo-cgi, line 34: I think there's a related story to update this to an Open button. If we do that we should consider either using AppButton or a new AppRowButton which we use for row button styling. Let's chat.

frontend/src/components/reports/ReportingHistoryTab.vue
```diff
@@ -1,90 +1,134 @@
...
+        </template>
+        <template #[`item.status`]="{ item }">
+          <span :class="getStatusClass(item)">{{ getStatusText(item) }}</span>
+        </template>
+        <template #[`item.submittedDate`]="{ item }">
+          {{ format.formatDate(item?.submittedDate) }}
+        </template>
+        <template #[`item.actions`]="{ item }">
+          <v-btn icon="mdi-eye-outline" variant="text" @click="viewSurveyResponse(item)" />
+          <v-btn icon="mdi-tray-arrow-down" variant="text" @click="false" />
```
weskubo-cgi, line 28: I think there's some desire to move away from icons to buttons/text. Maybe check the latest wireframe.

frontend/src/components/reports/PendingReportsTab.vue
```diff
@@ -5,73 +5,75 @@
...
+            <v-icon size="large">mdi-folder-open-outline</v-icon>
+          </v-btn>
+          <v-btn v-if="showTrash(item)" variant="text" @click="toggleCancelDialog(item)">
+            <v-icon size="large">mdi-trash-can-outline</v-icon>
```
weskubo-cgi, line 38: Icons should have an aria-label option for accessibility.  See FundingAgreementsTab.vue.

frontend/src/components/reports/ReportingSearchCard.vue
```diff
@@ -0,0 +1,213 @@
...
+                    </template>
+                  </v-list-item>
+                  <v-divider class="mt-2" />
+                </template>
+              </v-select>
+            </v-col>
+          </v-row> -->
+          <v-row>
+            <v-col cols="12" sm="6" lg="2" xl="3" class="pb-0">
+              <AppLabel>Status(s):</AppLabel>
```
weskubo-cgi, line 38: Status(s) looks a little odd since the plural of Status is Statuses.

backend/src/util/mapping/Mappings.js
```diff
@@ -366,25 +351,30 @@ const PermissionMappings = [
   { back: 'ofm_portal_privilege_number', front: 'permissionNumber' },
 ]
 
+// const ReportTemplateMappings = [
```
weskubo-cgi, line 354: I don't know that there's any need to comment this out given that it's not API code. Fine as is though.

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -51,7 +51,7 @@ export default {
         { title: 'Start Date', key: 'startDate' },
         { title: 'End Date', key: 'endDate' },
         { title: 'Status', key: 'status' },
```
weskubo-cgi, line 53: Not related to your change but it seems Status sort doesn't work.

frontend/src/components/reports/CancelSurveyResponseDialog.vue
```diff
@@ -23,36 +23,56 @@ import alertMixin from '@/mixins/alertMixin'
 
 export default {
   name: 'CancelSurveyResponseDialog',
+
```
weskubo-cgi, line 26: Was there a formatter change here? We don't normally bother with empty lines between declarations.

frontend/src/components/reports/PendingReportsTab.vue
```diff
@@ -5,73 +5,75 @@
...
 
-    <v-skeleton-loader :loading="processing" type="table-tbody">
-      <v-data-table :headers="headers" :items="filteredPendingReports" item-key="surveyResponseReferenceNumber" density="compact">
-        <template #[`item.alertType`]="{ item }">
-          <v-icon v-if="item.alertType === 'Due'" color="warning">mdi-alert</v-icon>
-          <v-icon v-else-if="item.alertType === 'Overdue'" color="error">mdi-alert-circle</v-icon>
-          <span class="pl-1">{{ item.alertType }}</span>
+    <v-skeleton-loader :loading="loading" type="table-tbody">
+      <AppAlertBanner v-if="isEmpty(pendingReports)" type="info">
+        <div>You are up to date with your monthly reports.</div>
```
weskubo-cgi, line 14: There's no need for a div here. The component already generates a div and p as a wrapper.

frontend/src/components/reports/PendingReportsTab.vue
```diff
@@ -5,73 +5,75 @@
...
+      </AppAlertBanner>
+      <v-data-table v-else :headers="headers" :items="filteredPendingReports" item-key="surveyTemplateId" density="compact" class="soft-outline">
+        <template #[`item.alert`]="{ item }">
+          <span v-if="isOverdue(item)">
```
weskubo-cgi, line 18: Span might be ok here but if you want to have an if-else without creating an extra wrapper you can use template as an invisible wrapper: https://vuejs.org/guide/essentials/conditional.html#v-if-on-template.

frontend/src/components/reports/ReportingHistoryTab.vue
```diff
@@ -1,90 +1,134 @@
...
-        <span :class="getStatusClass(item.status)" class="px-2 py-1">{{ item.status }}</span>
-      </template>
-      <template #[`item.actions`]="{ item }">
-        <v-btn icon="mdi-eye-outline" variant="text" @click="viewSurveyResponse(item)" />
-        <v-btn icon="mdi-tray-arrow-down" variant="text" @click="false" />
-      </template>
-    </v-data-table>
+    <v-skeleton-loader :loading="loading" type="table-tbody">
+      <AppAlertBanner v-if="isEmpty(submittedReports)" type="info" class="mt-4">
+        <div>You have no submitted reports.</div>
```
weskubo-cgi, line 14: Div not required.

frontend/src/components/reports/ReportingHistoryTab.vue
```diff
@@ -1,90 +1,134 @@
...
       }
+      return SURVEY_RESPONSE_STATUSES.COMPLETED
+      // TODO (vietle-cgi) - Add "Re-Submitted" status - pending on CRM team to add it.
+    },
+
+    getStatusClass(surveyResponse) {
+      if (surveyResponse?.isSubmittedLate) {
+        return 'status-red'
+      }
+      return 'status-green'
```
weskubo-cgi, line 131: Could use a ternary expression here.

frontend/src/components/reports/PendingReportsTab.vue
```diff
@@ -5,73 +5,75 @@
...
-      <v-data-table :headers="headers" :items="filteredPendingReports" item-key="surveyResponseReferenceNumber" density="compact">
-        <template #[`item.alertType`]="{ item }">
-          <v-icon v-if="item.alertType === 'Due'" color="warning">mdi-alert</v-icon>
-          <v-icon v-else-if="item.alertType === 'Overdue'" color="error">mdi-alert-circle</v-icon>
-          <span class="pl-1">{{ item.alertType }}</span>
+    <v-skeleton-loader :loading="loading" type="table-tbody">
+      <AppAlertBanner v-if="isEmpty(pendingReports)" type="info">
+        <div>You are up to date with your monthly reports.</div>
+      </AppAlertBanner>
+      <v-data-table v-else :headers="headers" :items="filteredPendingReports" item-key="surveyTemplateId" density="compact" class="soft-outline">
```
weskubo-cgi, line 16: We should have a mobile-breakpoint set.

weskubo-cgi (changes requested):

frontend/src/components/reports/CancelSurveyResponseDialog.vue
vietle-cgi: Agreed! we'll need to update the text of this box when we work on the delete ticket OFMCC-4126. I think we can use "restart your report", but I would like to get confirmation from Danna and the business.

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: Yes, I've let QA team know that those buttons will be updated in this ticket OFMCC-4146. I've tried use AppButton but the styling looks off, so I decided to update those icons to text in 4146.

frontend/src/components/reports/ReportingHistoryTab.vue
vietle-cgi: yes, it should be handled by this ticket: OFMCC-4146

frontend/src/components/reports/ReportingSearchCard.vue
vietle-cgi: Agreed! I copied how we do it in other screens. it means the user can select 1 or multiple statuses. Let me know if I should update it to Status(es)
image

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: sorry I'm not sure how it works in my case since I have v-if and v-else to show 2 icons

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: it is updated.

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: it is updated.

frontend/src/components/reports/ReportingHistoryTab.vue
vietle-cgi: it is updated.

frontend/src/components/reports/ReportingHistoryTab.vue
vietle-cgi: it is updated.

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: it is updated.

frontend/src/components/reports/CancelSurveyResponseDialog.vue
vietle-cgi: it is updated. sorry it was caused by my inconsistent preference. sometimes I want empty lines between sections, sometimes I don't

frontend/src/components/funding/FundingAgreementsTab.vue
vietle-cgi: it is fixed

backend/src/util/mapping/Mappings.js
vietle-cgi: let's keep it this way just in case we need to remove all of them in the future.

frontend/src/components/reports/PendingReportsTab.vue
weskubo-cgi: If you use <template v-if=""><template v-else> then it won't create an extra span holding your contents since template is an invisible wrapper. They're already contained in a td anyways. But if you want the span then it's fine.

frontend/src/components/reports/CancelSurveyResponseDialog.vue
weskubo-cgi: I hear ya.

frontend/src/components/reports/ReportingSearchCard.vue
weskubo-cgi: Status(es) would be more correct so let's update.

vietle-cgi: it's updated.

frontend/src/components/reports/PendingReportsTab.vue
vietle-cgi: I think I see what you mean. I've learned something new today. thank you!

weskubo-cgi (approved):

PR #235, opened 2024-06-12
ofmcc-1874 - update Application/Supp App status mapping

vietle-cgi:
In Application History page:
- For Core application, display Verified status as “In Review”
- If the user cancels a Core application from the Portal, that application will be deactivated in CRM and set to “Cancelled by SP” in CRM
- If the user cancels a Supplementary application from the Portal, that application will be deactivated in CRM and set to “Cancelled” in CRM

### DIFF
--- a/frontend/src/components/applications/CancelApplicationDialog.vue
+++ b/frontend/src/components/applications/CancelApplicationDialog.vue
@@ -20,7 +20,7 @@
 import AppButton from '@/components/ui/AppButton.vue'
 import AppDialog from '@/components/ui/AppDialog.vue'
 import alertMixin from '@/mixins/alertMixin'
-import { APPLICATION_STATUS_CODES, CRM_STATE_CODES } from '@/utils/constants'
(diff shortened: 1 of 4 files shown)


frontend/src/components/applications/CancelApplicationDialog.vue
```diff
@@ -68,7 +68,7 @@ export default {
       try {
         this.isLoading = true
         const payload = {
-          statusCode: APPLICATION_STATUS_CODES.CANCELLED,
+          statusCode: this.sourceApplicationType === 'Application' ? APPLICATION_STATUS_CODES.CANCELLED_BY_SP : SUPPLEMENTARY_APPLICATION_STATUS_CODES.CANCELLED,
```
weskubo-cgi, line 71: I noticed that the String 'Application' is now used twice in the file. Might be worth a constant.

weskubo-cgi (changes requested):

frontend/src/components/applications/CancelApplicationDialog.vue
vietle-cgi: it's updated

weskubo-cgi (approved):

PR #239, opened 2024-06-17
OFMCC-3706 Backend role validation

weskubo-cgi:
- Added permissions to all endpoints where defined (they were left off for simplicity's sake in the frontend but added to backend for completeness)
- Added role validation to all endpoints which don't clearly map to a permission (e.g. documents)
- Added one-off role validation for Impersonate since there's an endpoint which is only used during that functionality
- Minor code cleanup


frontend/src/main.js
```diff
@@ -38,8 +38,10 @@ async function loadLookupInfo() {
     await auth.getJwtToken()
     await app.getLookupInfo()
   } catch (e) {
-    // Flag errors that aren't 401 UNAUTHORIZED which is expected before the user is logged in
-    if (!e.response || e.response.status !== HttpStatus.UNAUTHORIZED) {
+    console.log(e)
```
vietle-cgi, line 41: do you want to keep this one?

vietle-cgi (changes requested):

frontend/src/main.js
weskubo-cgi: No. Removed.

vietle-cgi (approved):

PR #240, opened 2024-06-17
OFMCC-4143 - Unlock Closed Monthly Report

vietle-cgi:
Change is this PR:
- Refactor AppButton (simplify style and add btn-size props for small button)
- Add Unlock button to Reports table
- Add additional logics to NewRequestDialog for Reporting type.


frontend/src/components/ui/AppButton.vue
```diff
@@ -1,31 +1,44 @@
...
+import { isEmpty } from 'lodash'
+
 export default {
   inheritAttrs: true,
+  props: {
+    primary: {
+      type: Boolean,
+      default: true,
+    },
+    btnSize: {
```
weskubo-cgi, line 18: I believe we can simplify this by a) just reusing the existing size attribute and b) adding an additive style class to handle the sizes (e.g. small)

weskubo-cgi (changes requested):

frontend/src/components/ui/AppButton.vue
vietle-cgi: it's updated. thanks for your great suggestions!

weskubo-cgi (approved):

PR #251, opened 2024-06-20
feat: add fluentbit config for splunk and apply it to node

trev-dev:
Jira: OFMCC-641


tools/openshift/backend.dc.yaml
```diff
@@ -78,12 +72,56 @@ objects:
                 limits:
                   cpu: '${MAX_CPU}'
                   memory: '${MAX_MEM}'
+              volumeMounts:
```
weskubo-cgi, line 75: This volumeMounts element isn't the same position as backend-tls.dc.yaml which could be confusing if we have to compare or make change to both files.

tools/openshift/backend.dc.yaml
```diff
@@ -78,12 +72,56 @@ objects:
...
+                # - name: tls-certs
```
weskubo-cgi, line 76: Can we just remove tls-certs? They won't ever be needed as we have the -tls specific template.

weskubo-cgi (changes requested):

tools/openshift/backend.dc.yaml
trev-dev: Moved to the bottom of the image config in both files

tools/openshift/backend.dc.yaml
trev-dev: Yes :D

weskubo-cgi (approved):

PR #252, opened 2024-06-20
ofmcc-1773 - refactor inclusion policy update in Account Management

vietle-cgi:


frontend/src/components/organizations/OrganizationInfo.vue
```diff
@@ -242,45 +240,50 @@ export default {
...
       documentsToUpload: [],
       documentsToDelete: [],
-      showUploadDocumentsAlert: false,
       loadingInclusionPolicy: false,
     }
   },
-  async updated() {
+
+  computed: {
+    documentsExistOrBeingAdded() {
```
weskubo-cgi, line 248: Technically this doesn't need to be in computed as it's not referenced in the template. But it's also fine as is if you prefer to split this logic out.

weskubo-cgi (changes requested):

frontend/src/views/account-mgmt/ManageOrganizationView.vue
```diff
@@ -200,25 +200,26 @@ export default {
...
+          }
+          await OrganizationService.updateOrganization(this.organization?.organizationId, payload)
+          this.organization.hasInclusionPolicy = updatedHasInclusionPolicy
+        }
+        onSaveCompleteCallBack()
+        this.setSuccessAlert('Inclusion Policy updated successfully')
       } catch (error) {
-        this.setFailureAlert('Failed update Inclusion Policy on Organization: ' + this.organization.organizationId, error)
-        return
+        this.setFailureAlert('Failed update Inclusion Policy on Organization: ', error)
```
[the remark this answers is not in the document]
vietle-cgi: it's updated

frontend/src/components/organizations/OrganizationInfo.vue
vietle-cgi: let's keep it. I find it is easier to read this way.

weskubo-cgi (approved):

PR #259, opened 2024-06-25
fixed pdf icon showing when app still in draft status

jenbeckett:


frontend/src/views/applications/ApplicationsHistoryView.vue
```diff
@@ -88,20 +88,22 @@
...
-          </v-tooltip>
-          <v-btn v-else variant="text" @click="downloadPDF(item)">
-            <v-icon icon="fa:fa-regular fa-file-pdf"></v-icon>
-          </v-btn>
-
           <v-btn v-if="isApplicationCancellable(item)" variant="text" @click="toggleCancelDialog(item)">
             <v-icon icon="fa:fa-regular fa-trash-can"></v-icon>
           </v-btn>
+
+          <div v-else>
```
vietle-cgi, line 95: instead of v-else, I think we should have a separate computed value for showDownloadPDFButton. because if the isApplicationCancellable logic has Permission validation, I think your download button will display if the user has no permission.

vietle-cgi (changes requested):

frontend/src/views/applications/ApplicationsHistoryView.vue
jenbeckett: that's a good point.. I will add in a computed value that checks for View Applications permission

vietle-cgi (approved):

PR #260, opened 2024-06-25
update Funding Agreements to always return the newest FA or MOD

jenbeckett:
-- removed function getActiveFundingAgreementByFacilityId(facilityId) as it does not appear to be used anywhere 

always return the newest version (either the original Funding Agreement or MOD) that is active, unless being used for Supp Apps, in which case return the original FA. This is being updated here to support the backdate Transport App functionality (ofm 4772)

### DIFF
--- a/backend/src/components/fundingAgreements.js
+++ b/backend/src/components/fundingAgreements.js
@@ -10,12 +10,13 @@ const { isEmpty } = require('lodash')
 async function getFundingAgreements(req, res) {
   try {
     const fundingAgreements = []
-    let operation = 'ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode'
+    let operation =
+      'ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode,ofm_version_number'
     if (req.query?.includeEA) {
       operation += '&$expand=ofm_application($select=_ofm_expense_authority_value;$expand=ofm_expense_authority($select=ofm_first_name,ofm_last_name))'
     }
     const filter = `${buildDateFilterQuery(req?.query, 'ofm_start_date')}${buildFilterQuery(req?.query, FundingAgreementMappings)}`
-    operation += `&$filter=(${filter})`
+    operation += `&$filter=(${filter})&$orderby=ofm_version_number desc`
     const response = await getOperation(operation)
     response?.value?.forEach((funding) => {
       const fa = new MappableObjectForFront(funding, FundingAgreementMappings).toJSON()
--- a/backend/src/components/fundingAgreements.js
+++ b/backend/src/components/fundingAgreements.js
@@ -10,8 +10,7 @@ const { isEmpty } = require('lodash')
 async function getFundingAgreements(req, res) {
   try {
     const fundingAgreements = []
-    let operation =
-      'ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode,ofm_version_number'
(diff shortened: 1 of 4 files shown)


backend/src/components/fundingAgreements.js
```diff
@@ -10,12 +10,13 @@ const { isEmpty } = require('lodash')
 async function getFundingAgreements(req, res) {
   try {
     const fundingAgreements = []
-    let operation = 'ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode'
+    let operation =
+      'ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode,ofm_version_number'
```
vietle-cgi, line 14: Do we use "ofm_version_number" in frontend (because I don't see any mapping update)?

frontend/src/views/supp-allowances/SupplementaryAllowanceView.vue
```diff
@@ -140,15 +140,15 @@ export default {
         // navigate from the Application Confirmation page (after the providers have just submitted their core funding application)
         if (this.$route.params.applicationGuid) {
           this.application = await ApplicationService.getApplication(this.$route.params.applicationGuid)
-          this.application.fundingAgreement = await FundingAgreementService.getActiveFundingAgreementByApplicationId(this.$route.params.applicationGuid)
+          this.application.fundingAgreement = await FundingAgreementService.getActiveFundingAgreementByApplicationId(this.$route.params.applicationGuid, true)
```
vietle-cgi, line 143: should we add "true" flag to getActiveFundingAgreementByApplicationId in ApplicationsHistoryView? it's for enabling the Supplementary Application button

frontend/src/services/fundingAgreementService.js
```diff
@@ -47,19 +52,7 @@ export default {
...
-      return response?.data[0]
-    } catch (error) {
-      console.log(`Failed to get the active funding agreement by facility id - ${error}`)
-      throw error
-    }
-  },
-
-  // A facility can have only 1 Active funding agreement
+  //Node.js will order FA's so newest one (newest MOD agreement) will always be first
   async getActiveFundingAgreementByFacilityIdAndStatus(facilityId, statusCode) {
```
vietle-cgi, line 56: This function is used in SignFundingPopup. Do we need to update this function to return all FAs instead of just the 1st one? As you mentioned in the meeting, there could be multiple FA mod with signature pending?

vietle-cgi (changes requested):

backend/src/components/fundingAgreements.js
vietle-cgi: if we don't need it for now, we can remove it from the select

jenbeckett: no we don't use it - We just require that field so we can preform the "order by" operation

vietle-cgi: I think we can orderby without having it in select statement

frontend/src/services/fundingAgreementService.js
jenbeckett: I was on the fence about this one.. It sounded like there "shouldn't be" the case of multiple FA's to sign.. However even if there is a FA to sign, and it is not the most recent one - this will still return it. I can def change it to support that... what do you think?

backend/src/components/fundingAgreements.js
jenbeckett: hmm you are right! I didn't think that was the case. Ok ok it's gone :)

frontend/src/services/fundingAgreementService.js
vietle-cgi: I think yours looks good because we only show max 3 FAs in the popup. also, once they go to Funding Agreement Tab, they can see the full list of all Sig Pending FAs. we should be good.

frontend/src/views/supp-allowances/SupplementaryAllowanceView.vue
jenbeckett: no we don't need to - the check that happens in ApplicationService looks at the endDate to confirm if the FA (or MOD) is active. End dates do not change across FA and MOD's . 

We re-load the FA from Dynamics on the SuppApp page anyways to get the original FA so I think we are okay here.

vietle-cgi (approved):

PR #268, opened 2024-06-28
Ofmcc 4322 unexpected expense

jenbeckett:
Create an Irregular Expense request 

Note - I had to load the applications (and funding agreements) - from Messaging View. The Dialog did not populate correctly if the data was not already pre-loaded. This also saves on another Dynamics call when starting the request from the Applications View page


frontend/src/views/applications/ApplicationsHistoryView.vue
```diff
@@ -109,6 +128,11 @@
     </v-skeleton-loader>
     <CancelApplicationDialog :show="showCancelDialog" :applicationId="cancelledApplicationId" :applicationType="applicationTypeToCancel" @close="toggleCancelDialog" @cancel="cancelApplication" />
     <AppBackButton id="back-home-button" width="220px" :to="{ name: 'home' }">Home</AppBackButton>
+    <NewRequestDialog
+      class="pa-0"
```
weskubo-cgi, line 132: I see this class in a few places but it doesn't seem to impact the layout at all so it can be removed.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -65,13 +65,13 @@
...
               <v-select
-                v-if="isAnAccountMaintenanceRequest"
+                v-if="isAnAccountMaintenanceRequest || isAnIrregularExpenseRequest"
                 id="selectFacility"
                 v-model="newRequestModel.facilities"
                 placeholder="[select facility]"
                 variant="outlined"
                 density="compact"
-                :items="facilities"
+                :items="isAnIrregularExpenseRequest ? filterFacilitiesWithoutFA : facilities"
```
weskubo-cgi, line 74: If you made this a computed method (filteredFacilities) it would make this attribute simpler and you wouldn't need the additional filterFacilitiesWithoutFA method as the logic could just be in there. I.e. if not irregular return facilties else filter by fa.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -230,6 +230,18 @@
                 :disabled="isLoadingOrDisabled"></v-textarea>
             </v-col>
           </v-row>
+
+          <v-row v-if="isAnIrregularExpenseRequest" class="my-3">
+            <v-col cols="12">
+              <h5>Download a Irregular Expense Form then fill it out!</h5>
```
weskubo-cgi, line 236: Is this the confirmed text or are we waiting for wording?

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -230,6 +230,18 @@
...
+            </v-col>
+            <v-row>
+              <v-col cols="12" class="ml-3">
+                <AppButton id="download-irregular-expense-form" size="large" width="200px" :disabled="isDisabled" :loading="isLoading">Download Form</AppButton>
```
weskubo-cgi, line 240: My understanding is that this would be a hyperlink to an externally hosted site.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -395,6 +417,9 @@ export default {
     isAnAccountMaintenanceRequest() {
       return this.newRequestModel.requestCategoryId === this.getRequestCategoryIdByName(REQUEST_CATEGORY_NAMES.ACCOUNT_MAINTENANCE)
     },
+    isAnIrregularExpenseRequest() {
```
weskubo-cgi, line 420: Let's drop the An from the computed names.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -467,6 +492,10 @@ export default {
           'A phone/cell or email is required when Facility phone/email checked',
       ]
     },
+
+    filterFacilitiesWithoutFA() {
+      return this.facilities.filter((fac) => this.fundingAgreements.find((app) => app.facilityId == fac.facilityId))
```
weskubo-cgi, line 497: I think you can use some instead of find. Also, you probably want to use ===.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -493,32 +522,29 @@ export default {
...
     },
   },
-  created() {
+  async created() {
     this.PHONE_FORMAT = PHONE_FORMAT
     this.EMAIL_FORMAT = EMAIL_FORMAT
     this.AUTO_REPLY_MESSAGE = 'Your change request is complete.'
     this.setUpDefaultNewRequestModel()
+
+    if (this.PERMISSIONS.SUBMIT_CHANGE_REQUEST) {
```
weskubo-cgi, line 545: This condition is checking against a constant so it will always return true.

frontend/src/components/messages/NewRequestDialog.vue
```diff
@@ -493,32 +522,29 @@ export default {
...
+      this.isOFMValid = await this.validateOfmProgram()
```
weskubo-cgi, line 546: I believe the validateOfmProgram check depends on the selection of facilities which hasn't been done at this point. The best place might be on a watcher for subCategories.

weskubo-cgi (changes requested):

weskubo-cgi (changes requested):

PR #269, opened 2024-07-03
Ofmcc 3478 cr refresh issue

weskubo-cgi:
Updated to refresh on submit not on close confirmation.

### DIFF
--- a/frontend/src/views/account-mgmt/ManageFacilityView.vue
+++ b/frontend/src/views/account-mgmt/ManageFacilityView.vue
@@ -140,7 +140,7 @@
       :defaultRequestCategoryId="getRequestCategoryIdByName(REQUEST_CATEGORY_NAMES.ACCOUNT_MAINTENANCE)"
       :defaultFacility="facility"
       @close="toggleChangeRequestDialog"
-      @close-confirmation="handleCloseConfirmation" />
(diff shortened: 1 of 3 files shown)


frontend/src/views/account-mgmt/ManageFacilityView.vue
```diff
@@ -140,7 +140,7 @@
       :defaultRequestCategoryId="getRequestCategoryIdByName(REQUEST_CATEGORY_NAMES.ACCOUNT_MAINTENANCE)"
       :defaultFacility="facility"
       @close="toggleChangeRequestDialog"
-      @close-confirmation="handleCloseConfirmation" />
+      @submit-phone-email="handleCRSubmit" />
```
jenbeckett, line 143: here you have the event in kebab case, but the emit is done in camelCase. It still works regardless, but is it better to have them match? Just curious because it seems like there are multiple ways we name emits/events in the project

weskubo-cgi: It's best to follow https://vuejs.org/guide/components/events.html which borrows from https://vuejs.org/guide/components/props.html#prop-name-casing. The project is admittedly a mix.

jenbeckett (approved):

PR #277, opened 2024-07-09
Ofmcc 638 setup prod

weskubo-cgi:


tools/openshift/frontend-tls.dc.yaml
```diff
@@ -0,0 +1,222 @@
...
+    required: false
+  - name: TDAD_CONTACT_EMAIL
+    description: The contact email for 10 dollar a day daycare centers
+    required: false
+  - name: CCOF_URL
+    description: The URL for the CCOF portal that belongs to the same application environment.
+    required: false
+  - name: IRREGULAR_EXPENSE_FORM_URL
+    description: The URL to download the Irregular Expense application form.
+    required: false
```
SoLetsDev, line 213: Seems like these should be set to true?

SoLetsDev (approved):

PR #281, opened 2024-07-09
Add Approved Supp Apps to Funding Table

jenbeckett:
We do not have the notification PDF to be displayed on this screen yet - so I hid the 'open' buttons for applications that are not Base Funding.

### DIFF
--- a/frontend/src/components/funding/FundingAgreementsTab.vue
+++ b/frontend/src/components/funding/FundingAgreementsTab.vue
@@ -32,9 +32,10 @@ import FundingSearchCard from '@/components/funding/FundingSearchCard.vue'
 import alertMixin from '@/mixins/alertMixin.js'
 import { useAuthStore } from '@/stores/auth'
 import FundingAgreementService from '@/services/fundingAgreementService'
-import { FUNDING_AGREEMENT_STATUS_CODES, SUPPLEMENTARY_APPLICATION_STATUS_CODES } from '@/utils/constants'
+import { FUNDING_AGREEMENT_STATUS_CODES, SUPPLEMENTARY_APPLICATION_STATUS_CODES, BLANK_FIELD } from '@/utils/constants'
 import format from '@/utils/format'
 
+const BASE_FUNDING = 'Base Funding'
 export default {
   name: 'FundingAgreementsTab',
   components: { AppButton, FundingSearchCard },
@@ -61,7 +62,6 @@ export default {
   },
 
   async created() {
-    this.BASE_FUNDING = 'Base Funding'
     this.loading = true
     this.format = format
     this.FUNDING_AGREEMENT_STATUS_CODES = FUNDING_AGREEMENT_STATUS_CODES
@@ -87,7 +87,7 @@ export default {
             endDate: app.endDate,
             fundingAgreementNumber: app.supplementaryReferenceNumber,
             fundingAgreementType: app.supplementaryTypeDescription,
-            expenseAuthority: '- - - -',
+            expenseAuthority: this.BLANK_FIELD,
             facilityName: applications.find((el) => app.applicationId === el.applicationId).facilityName,
             statusCode: FUNDING_AGREEMENT_STATUS_CODES.ACTIVE, //use the FA code to highlight the statusName in green
             statusName: app.supplementaryApplicationStatus,
@@ -106,7 +106,7 @@ export default {
             const facilityFas = await FundingAgreementService.getFAsByFacilityIdAndStartDate(facility.facilityId, searchQueries?.dateFrom, searchQueries?.dateTo)
             if (facilityFas) {
               facilityFas.forEach((fa) => {
-                fa.fundingAgreementType = this.BASE_FUNDING
+                fa.fundingAgreementType = BASE_FUNDING
                 fa.priority = fa.statusCode === FUNDING_AGREEMENT_STATUS_CODES.SIGNATURE_PENDING ? 1 : 0
               })
               this.fundingAgreements.push(...facilityFas)
@@ -125,7 +125,7 @@ export default {
 
     showOpen(item) {
       //TODO : add in link for PDF -- we don't have a PDF to pull for supp apps yet
-      if (item.fundingAgreementType !== this.BASE_FUNDING) {
+      if (item.fundingAgreementType !== BASE_FUNDING) {
         return false
       }
       return (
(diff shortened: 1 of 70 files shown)


frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +60,43 @@ export default {
     ...mapState(useAuthStore, ['userInfo']),
   },
 
-  created() {
+  async created() {
+    this.BASE_FUNDING = 'Base Funding'
+    this.loading = true
     this.format = format
     this.FUNDING_AGREEMENT_STATUS_CODES = FUNDING_AGREEMENT_STATUS_CODES
```
weskubo-cgi, line 67: If constants aren't accessed in the template, they don't need to be redefined in created().

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +60,43 @@ export default {
     ...mapState(useAuthStore, ['userInfo']),
   },
 
-  created() {
+  async created() {
+    this.BASE_FUNDING = 'Base Funding'
```
weskubo-cgi, line 64: This constant isn't used in the template, so it could just be defined like a regular js constant, before export default.

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +60,43 @@ export default {
...
+          )
+        ).flat()
+
+        this.supplementaryApplications.forEach((app) => {
+          this.fundingAgreements.push({
+            startDate: app.startDate,
+            endDate: app.endDate,
+            fundingAgreementNumber: app.supplementaryReferenceNumber,
+            fundingAgreementType: app.supplementaryTypeDescription,
+            expenseAuthority: '- - - -',
```
weskubo-cgi, line 90: This is available as a constant in constants.js.

weskubo-cgi (dismissed):

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +61,42 @@ export default {
...
     this.FUNDING_AGREEMENT_STATUS_CODES = FUNDING_AGREEMENT_STATUS_CODES
+    await this.loadApprovedSuppApps()
+    this.loading = false
   },
 
   methods: {
+    async loadApprovedSuppApps() {
+      try {
+        this.loading = true
+        const applications = await ApplicationService.getActiveApplications()
```
vietle-cgi, line 76: the Funding tab can display all Historical Funding Agreement (expired). Should we also display historical supplementary funding?

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -84,19 +116,21 @@ export default {
         this.fundingAgreements?.sort((a, b) => b.priority - a.priority) // FA Signature Pending status at the top
       } catch (error) {
         this.setFailureAlert('Failed to load funding agreements', error)
-      } finally {
```
vietle-cgi, line 87: there is a Search card in this view. if we remove loading = false from this function, the screen will stuck at loading when the user click search button. I think we should keep this finally part

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +61,42 @@ export default {
...
 
   methods: {
+    async loadApprovedSuppApps() {
+      try {
+        this.loading = true
+        const applications = await ApplicationService.getActiveApplications()
+
+        this.supplementaryApplications = (
+          await Promise.all(
+            applications.map((application) => ApplicationService.getSupplementaryApplications(application.applicationId, `statusCode=${SUPPLEMENTARY_APPLICATION_STATUS_CODES.APPROVED}`)),
```
vietle-cgi, line 80: There is a search card in this screen, and I don't think it works well with your Supp Funding.

vietle-cgi (changes requested):

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +60,43 @@ export default {
...
+        ).flat()
+
+        this.supplementaryApplications.forEach((app) => {
+          this.fundingAgreements.push({
+            startDate: app.startDate,
+            endDate: app.endDate,
+            fundingAgreementNumber: app.supplementaryReferenceNumber,
+            fundingAgreementType: app.supplementaryTypeDescription,
+            expenseAuthority: '- - - -',
+            facilityName: applications.find((el) => app.applicationId === el.applicationId).facilityName,
```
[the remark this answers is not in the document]
jenbeckett: done

frontend/src/components/funding/FundingAgreementsTab.vue
jenbeckett: should work better now !

vietle-cgi (changes requested):

frontend/src/components/funding/FundingAgreementsTab.vue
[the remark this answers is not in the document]
vietle-cgi: I don't see this change? we need it just in case applications.find return null

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +61,44 @@ export default {
     ...mapState(useAuthStore, ['userInfo']),
   },
 
-  created() {
+  async created() {
```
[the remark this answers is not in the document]
jenbeckett: nope - it's gone

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -59,12 +61,44 @@ export default {
...
     this.format = format
     this.FUNDING_AGREEMENT_STATUS_CODES = FUNDING_AGREEMENT_STATUS_CODES
+
+    this.loading = false
   },
 
   methods: {
+    async loadApprovedSuppApps(startDateFrom, startDateTo) {
+      try {
+        this.loading = true
```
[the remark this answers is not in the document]
jenbeckett: removed

frontend/src/components/funding/FundingAgreementsTab.vue
jenbeckett: I don't think we have the functionality to support expired yet - I will write a note though for it to be picked up in future work

vietle-cgi (approved):

PR #289, opened 2024-07-17
Ofmcc 5049 ccsea union questions

vietle-cgi:
Changes in this PR:
- OFMCC-5049 - add union and cssea questions to Application
- OFMCC-5252 - update Application's declaration wording
- OFMCC-5115 - fix scroll behavior when navigating between Application's pages

### DIFF
--- a/frontend/src/utils/rules.js
+++ b/frontend/src/utils/rules.js
@@ -7,6 +7,8 @@ const rules = {
         return true
       } else if (Array.isArray(v)) {
         return v.length > 0 || REQUIRED_MSG
+      } else if (typeof v === 'boolean') {
+        return v != null
       } else if (!v) {
         return REQUIRED_MSG
       }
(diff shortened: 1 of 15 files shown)


frontend/src/utils/rules.js
```diff
@@ -7,6 +7,8 @@ const rules = {
         return true
       } else if (Array.isArray(v)) {
         return v.length > 0 || REQUIRED_MSG
+      } else if (typeof v === 'boolean') {
+        return v != null
```
jenbeckett, line 11: I think this would always return true?

vietle-cgi: agreed. I'll move it to the 1st condition as below:
```
if (v === 0 || typeof v === 'boolean') {
        return true
      }
```

vietle-cgi: it's updated!

jenbeckett (approved):

PR #292, opened 2024-07-19
ofmcc-5160 - add due-overdue monthly report banner

vietle-cgi:


frontend/src/views/HomeView.vue
```diff
@@ -1,4 +1,15 @@
 <template>
+  <v-row v-if="showReportsAlertBanner" no-gutters class="blue-background justify-center pa-2">
+    <v-banner bg-color="#f9f1c6" :lines="$vuetify.display.smAndUp ? 'one' : 'two'" max-width="700px" rounded class="alert-banner">
```
weskubo-cgi, line 3: Not sure if it's worth doing anything about but there is a resolution where the text get cuts off before it wraps to two lines. If you switch to mdAndUp it wraps pretty early.
image

weskubo-cgi (changes requested):

frontend/src/views/HomeView.vue
vietle-cgi: yes, I've also realized this when issue when I implemented it. I've tried different way and decided to do it this way because it only happens to some specific resolutions (600 x ... -> 644 x ...).

weskubo-cgi (approved):

PR #307, opened 2024-07-26
OFMCC-4619 - Display system message

vietle-cgi:
Add public endpoint to backend. This public endpoint is accessible without any authentication


frontend/src/views/LoginView.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <AppHeroImage />
   <v-container class="login-container" fluid v-bind="$attrs">
+    <AppAlertBanner v-for="message in systemMessages" type="info">{{ message.content }}</AppAlertBanner>
```
weskubo-cgi, line 4: This v-for requires a key. I don't think there's a natural key so you could add a guid after retrieval.

backend/src/components/public.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
```
weskubo-cgi, line 1: While I love the idea of having a public route to make it clear what functionality is being exposed publicly I think the components can still be aligned along functional/business concerns. While it's not exactly "lookup" data it does do caching you could certainly lump this into lookup.js or rename this component to be just for system messages.

backend/src/routes/public.js
```diff
@@ -0,0 +1,9 @@
+const express = require('express')
+const router = express.Router()
+const { getSystemMessages } = require('../components/public')
+
+router.get('/system-messages', (req, res) => {
```
weskubo-cgi, line 5: My only concern with a public route that hits Dynamics is the potential to abuse it via DOS attacks. Now, the risk of overwhelming Dynamics is really minimal since we are caching data anyways. But I think this might be a good opportunity to add rate limiting (https://www.npmjs.com/package/express-rate-limit) support to the backend, specifically to this endpoint. Since we have a distributed backend you would also need https://www.npmjs.com/package/rate-limit-redis. We can probably add a separate NFR ticket for this for one of us to work on.

frontend/src/services/publicService.js
```diff
@@ -0,0 +1,14 @@
+import ApiService from '@/common/apiService'
```
weskubo-cgi, line 1: As for backend components, let's align the services functionally.

weskubo-cgi (changes requested):

frontend/src/services/publicService.js
vietle-cgi: do you mean: instead of having publicService, we should just add getSystemMessages function to lookupService --> match with backend component?

weskubo-cgi: Yes, if you've utilized lookup.js in the backend, let's use lookupService in the frontend.

frontend/src/views/LoginView.vue
vietle-cgi: it's updated

backend/src/components/public.js
vietle-cgi: it's updated

backend/src/routes/public.js
vietle-cgi: I've added this rate limit to app.js. could you please have a look? I'm using the recommended setting in the express-rate-limit document.

frontend/src/services/publicService.js
vietle-cgi: it's updated

weskubo-cgi (approved):

PR #316, opened 2024-08-08
Add EFX environment

trev-dev:
Jira: OFMCC-637

### DIFF
--- a/.github/workflows/deploy-to-openshift-frontend-efx.yml
+++ b/.github/workflows/deploy-to-openshift-frontend-efx.yml
@@ -1,0 +1,159 @@
+name: 4 EFX - Deploy Frontend
+
+env:
+  # 🖊️ EDIT your repository secrets to log into your OpenShift cluster and set up the context.
+  # See https://github.com/redhat-actions/oc-login#readme for how to retrieve these values.
+  # To get a permanent token, refer to https://github.com/redhat-actions/oc-login/wiki/Using-a-Service-Account-for-GitHub-Actions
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy-to-openshift-frontend-efx.yml
```diff
@@ -0,0 +1,159 @@
...
+  MAX_MEM: '250Mi'
+  # SITE_URL should have no scheme or port. It will be prepended with https://
+  HOST_ROUTE: ${{ secrets.SITE_URL }}
+  CA_CERT: ${{ secrets.CA_CERT }}
+  CERTIFICATE: ${{ secrets.CERTIFICATE }}
+  PRIVATE_KEY: ${{ secrets.PRIVATE_KEY }}
+
+  # Frontend configuration
+  BANNER_ENVIRONMENT: 'EFX'
+  BANNER_COLOR: '#800080'
```
weskubo-cgi, line 45: I have no idea where the banner colors came from but we need a new one. Maybe a darker green (Dev is light green) as blue/yellow conflict with the palette and red screams error.

weskubo-cgi (changes requested):

.github/workflows/deploy-to-openshift-frontend-efx.yml
trev-dev: @weskubo-cgi 
image

How's this?

weskubo-cgi (approved):

PR #318, opened 2024-08-15
Ofmcc 2105 check for inclusion policy

jenbeckett:


frontend/src/views/supp-allowances/SupplementarySubmitView.vue
```diff
@@ -228,6 +238,7 @@ export default {
     await this.loadData()
   },
   methods: {
```
vietle-cgi, line 240: this issue is not related to this PR. just something I realize while I test this PR. For Impersonate or user with readonly access, I think we should hide the "Cancel" button and we should have a readonly check for saveApplication function (i.e.: we should not trigger saveApplication if it is readonly - in both SupplementaryFormView and SupplementarySubmitView): 
image

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
```diff
@@ -157,7 +165,9 @@ export default {
     },
     isSupportComplete() {
       const model = this.getModel(SUPPLEMENTARY_TYPES.SUPPORT)
-      if (!model?.supportFundingModel.includes(this.SUPPORT_CHECKBOX_LABELS.find((item) => item.label === 'Other').value)) {
+      if (!this.currentOrg.hasInclusionPolicy && model) {
```
vietle-cgi, line 168: I think a warning banner is good enough for missing inclusion policy. The "missing required info" alert should be only for the incomplete Supplementary form (not for missing inclusion policy)
image

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
```diff
@@ -236,6 +247,9 @@ export default {
       try {
         this.loading = true
         this.$emit('process', true)
+        if (!this.currentOrg) {
```
vietle-cgi, line 250: If the provider updates their inclusion policy to Yes but they don't refresh the page, the inclusion policy in Org store will not be updated (e.g.: still set to False) --> I think we should always reload getOrganizationInfo().

Also, I think we should call getOrganizationInfo() in SupplementaryAllowanceView instead of calling it twice in SupplementaryFormView and SupplementarySubmitView.

vietle-cgi (changes requested):

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
jenbeckett: good point I will change that

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
jenbeckett: done!

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
vietle-cgi: When I click "Back" from the Submit page (Impersonate), I'm still see this error (403 Forbidden for patch request)

frontend/src/views/supp-allowances/SupplementaryAllowanceView.vue
```diff
@@ -99,12 +100,13 @@ export default {
   },
 
   computed: {
+    ...mapState(useOrgStore, ['currentOrg']),
```
vietle-cgi, line 103: I don't think it is used?

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
```diff
@@ -174,16 +180,24 @@ export default {
...
+      return this.isIndigenousComplete && this.isSupportComplete && this.isTransportComplete && this.hasSupportApplicationAndInclusionPolicy
     },
     readonly() {
       return !this.isApplicationComplete || this.processing || this.loading || !this.hasPermission(this.PERMISSIONS.APPLY_FOR_FUNDING)
     },
+    //this function disables submit if user has filled out a Support Needs App - but then after removes inclusion policy.
+    //The logic was moved out of isSupportComplete() so the red 'missing info' messaging won't appear when user is missing inclusion policy
+    hasSupportApplicationAndInclusionPolicy() {
+      if (!this.getModel(SUPPLEMENTARY_TYPES.SUPPORT)) return true
+      return this.currentOrg.hasInclusionPolicy
```
vietle-cgi, line 192: I think it could be simplified to 1 line? (i.e.: return !this.getModel(SUPPLEMENTARY_TYPES.SUPPORT) || this.currentOrg.hasInclusionPolicy)

vietle-cgi (changes requested):

frontend/src/views/supp-allowances/SupplementaryAllowanceView.vue
jenbeckett: done

frontend/src/views/supp-allowances/SupplementarySubmitView.vue
jenbeckett: done

vietle-cgi (approved):

PR #338, opened 2024-08-26
small ui fixes and reply button replies again

jenbeckett:
Changed message button CSS on focus to not have the tiny bit of blue background 
Added back in the "Reply" button functionality that was mistakenly removed from a previous PR
Added <span> tags on messaging buttons, so whitespace would no longer be underlined 

I left 'Notifications' and 'Messages' tab words bold - as I think they look better but open to suggestions on this

### DIFF
--- a/frontend/src/components/messages/MessagesTab.vue
+++ b/frontend/src/components/messages/MessagesTab.vue
@@ -4,15 +4,15 @@
       <v-row class="mt-0 ml-0">
         <v-col cols="12">
           <AppButton size="small" class="messages-button" @click="toggleNewRequestDialog()">
-            <v-icon left>mdi-email-plus-outline</v-icon>
+            <v-icon class="icon mr-1" left>mdi-email-plus-outline</v-icon>
             <span>New message</span>
           </AppButton>
           <AppButton size="small" class="messages-button mx-1" :primary="false" @click="toggleMarkUnreadButtonInMessageTable()">
-            <v-icon left>mdi-email-outline</v-icon>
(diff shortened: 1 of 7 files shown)


frontend/src/components/messages/MessagesTab.vue
```diff
@@ -5,15 +5,15 @@
         <v-col cols="12">
           <AppButton size="small" class="messages-button" @click="toggleNewRequestDialog()">
             <v-icon left>mdi-email-plus-outline</v-icon>
-            New message
+            <span>New message</span>
```
vietle-cgi, line 8: Hi Jen, could we remove the span tag in Notification's buttons as well so they will look consistent?

frontend/src/components/messages/MessagesTab.vue
```diff
@@ -91,4 +91,9 @@ export default {
 .messages-button {
   display: inline-block;
 }
+
+.messages-button:focus {
```
vietle-cgi, line 95: Hi Jen, could we add this style for Notification's buttons as well so they will look consistent?

vietle-cgi (changes requested):

frontend/src/components/messages/MessagesTab.vue
jenbeckett: done :)

frontend/src/components/messages/MessagesTab.vue
jenbeckett: done :)

vietle-cgi (approved):

PR #347, opened 2024-08-29
Ofmcc 5789 transport validation enhancement

jenbeckett:
Changed the messaging in 'rules' for the min value -- currently this PR is the only files that use that error message, so I don't think there should be an impact there 

Prevent save / next if user enters '0' on some fields on Transport App. I purposefully did not set the 'min' value on format to 1 - because it becomes annoying if a user wants to type in a different number that does not start with one. 

Works with both current and next year scenarios

### DIFF
--- a/frontend/src/components/supp-allowances/TransportationAllowance.vue
+++ b/frontend/src/components/supp-allowances/TransportationAllowance.vue
@@ -83,7 +83,7 @@
                 <AppNumberInput
                   v-model="model.odometer"
                   required
-                  :format="wholeNumberNonZeroFormat"
+                  :format="wholeNumberFormat"
                   suffix="km"
                   variant="outlined"
                   density="compact"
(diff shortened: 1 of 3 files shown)


frontend/src/components/supp-allowances/TransportationAllowance.vue
```diff
@@ -80,15 +80,15 @@
...
               </v-col>
               <v-col cols="6" xl="7" class="pt-2 text-center">
-                <v-text-field
+                <AppNumberInput
                   v-model="model.odometer"
                   required
-                  type="number"
+                  :format="wholeNumberFormat"
                   suffix="km"
                   variant="outlined"
```
vietle-cgi, line 88: I think we don't need "required", "variant", and "density".

frontend/src/components/supp-allowances/TransportationAllowance.vue
```diff
@@ -98,15 +98,15 @@
                 <p>Estimated Yearly KM:</p>
               </v-col>
               <v-col cols="6" xl="7" class="pt-2 text-center">
-                <v-text-field
+                <AppNumberInput
                   v-model="model.estimatedMileage"
                   required
```
vietle-cgi, line 103: I think we don't need "required", "variant", and "density".

vietle-cgi (changes requested):

vietle-cgi (approved):

PR #352, opened 2024-09-06
OFMCC-5293 - Application Eligibility

vietle-cgi:
Add Application Eligibility and some validations
UI update - application header

### DIFF
--- a/frontend/src/components/applications/ApplicationNavBar.vue
+++ b/frontend/src/components/applications/ApplicationNavBar.vue
@@ -145,22 +145,15 @@ export default {
       if (this.isSelectFacilityPage) {
         return 'mdi-circle'
       }
-      switch (item.routeName) {
-        case APPLICATION_ROUTES.FACILITY_DETAILS:
-          return this.isFacilityDetailsComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.ELIGIBILITY:
-          return this.isEligibilityComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.SERVICE_DELIVERY:
-          return this.isServiceDeliveryComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.OPERATING_COSTS:
-          return this.isOperatingCostsComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.STAFFING:
-          return this.isStaffingComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.REVIEW:
-          return this.isApplicationComplete ? 'mdi-check-circle' : 'mdi-circle'
-        case APPLICATION_ROUTES.SUBMIT:
-          return this.isDeclareSubmitComplete ? 'mdi-check-circle' : 'mdi-circle'
-      }
+      const isComplete =
+        (this.isRouteNameEqual(item, APPLICATION_ROUTES.FACILITY_DETAILS) && this.isFacilityDetailsComplete) ||
+        (this.isRouteNameEqual(item, APPLICATION_ROUTES.ELIGIBILITY) && this.isEligibilityComplete) ||
+        (this.isRouteNameEqual(item, APPLICATION_ROUTES.SERVICE_DELIVERY) && this.isServiceDeliveryComplete) ||
+        (this.isRouteNameEqual(item, APPLICATION_ROUTES.OPERATING_COSTS) && this.isOperatingCostsComplete) ||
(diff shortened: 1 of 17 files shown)


frontend/src/components/applications/EligibilityNotificationDialog.vue
```diff
@@ -0,0 +1,81 @@
...
+      <div class="confirm-dialog-text">
+        Thank you for your interest in joining the $10 a Day ChildCareBC Program. Based on the answers you have provided, your facility has not met the minimum eligibility criteria to participate at
+        this time. For more information about the $10 a Day ChildCareBC Program, please visit our
+        <a href="https://www2.gov.bc.ca/gov/content?id=F0D0FD2A00064CDDBF634CF83E4E2599" target="_blank">website</a>
+      </div>
+    </template>
+    <template #button>
+      <v-row justify="space-around">
+        <v-col cols="12" md="6" class="d-flex justify-center">
+          <AppButton id="dialog-cancel-application" :primary="false" size="large" min-width="250px" max-width="450px" :loading="isLoading" @click="cancel">Cancel Application</AppButton>
```
jenbeckett, line 13: After clicking No on one of the knockout questions - I saw the popup then clicked 'cancel' on the popup, as prompted. However - after clicking cancel, I was still able to close the popup, change my answers all to yes - and then proceed to the end of the application, where I could not submit (assuming because there is no application left)? But I could see how that would maybe be confusing for the user. 
 
Just thinking maybe we should do something after a user clicks cancel - send them back to the home page maybe?

frontend/src/views/applications/EligibilityView.vue
```diff
@@ -0,0 +1,284 @@
...
+          id="health-authority-good-standing"
+          v-model="model.healthAuthorityGoodStanding"
+          :rules="rules.required"
+          :hide-details="readonly"
+          :disabled="readonly"
+          @update:model-value="validateResponse(model.healthAuthorityGoodStanding)" />
+      </div>
+      <div>
+        <div>
+          Do all your current Early Childhood Educator(s) have an active, valid certificate? (If your licence does not require an Early Childhood Educator and you have none working in your operation,
```
jenbeckett, line 51: When I try and select a N/A field - my application does not save

image

I think this is because N/A is not a value in CRM?

frontend/src/components/applications/EligibilityNotificationDialog.vue
vietle-cgi: good point! I will update it

frontend/src/components/applications/ApplicationNavBar.vue
```diff
@@ -136,6 +148,8 @@ export default {
       switch (item.routeName) {
         case APPLICATION_ROUTES.FACILITY_DETAILS:
           return this.isFacilityDetailsComplete ? 'mdi-check-circle' : 'mdi-circle'
+        case APPLICATION_ROUTES.ELIGIBILITY:
+          return this.isEligibilityComplete ? 'mdi-check-circle' : 'mdi-circle'
```
weskubo-cgi, line 152: Not sure if you want to refactor the logic to get the correct icon into it's own function? It's always the same pair of icons so it would save some code. I know you like less code!

frontend/src/components/applications/EligibilityNotificationDialog.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <AppDialog v-model="isDisplayed" title="Eligibility Notification" :is-loading="isLoading" persistent max-width="55%" @close="closeDialog">
+    <template #content>
+      <div class="confirm-dialog-text">
+        Thank you for your interest in joining the $10 a Day ChildCareBC Program. Based on the answers you have provided, your facility has not met the minimum eligibility criteria to participate at
+        this time. For more information about the $10 a Day ChildCareBC Program, please visit our
+        <a href="https://www2.gov.bc.ca/gov/content?id=F0D0FD2A00064CDDBF634CF83E4E2599" target="_blank">website</a>
```
weskubo-cgi, line 7: This is probably a mockup issue but should there be a period after the website link?

frontend/src/components/applications/EligibilityNotificationDialog.vue
vietle-cgi: it's updated. thanks Jen!

weskubo-cgi (changes requested):

frontend/src/components/applications/EligibilityNotificationDialog.vue
vietle-cgi: it's updated

frontend/src/components/applications/ApplicationNavBar.vue
vietle-cgi: yes, I love clean code! I think it looks cleaner now. thanks Wes!

weskubo-cgi (approved):

PR #359, opened 2024-09-13
ofmcc-2365 - filter out inactive facilities

vietle-cgi:


frontend/src/views/applications/ApplicationsHistoryView.vue
```diff
@@ -235,7 +235,9 @@ export default {
       }
     },
     isAddCoreApplicationAllowed() {
-      return this.userInfo?.facilities?.some((facility) => facility.intakeWindowCheckForAddApplication && facility.ccofEnrolmentCheckForAddApplication)
+      return this.userInfo?.facilities?.some(
```
jenbeckett, line 238: would it be worth adding a check here for a re-directed application? The facility is still marked as active - but they shouldn't be allowed to start another OFM core app

vietle-cgi: thanks Jen! it's updated

weskubo-cgi (approved):

PR #369, opened 2024-09-20
OFMCC-6066 - add Multiple program and update some CCOF validations

vietle-cgi:


backend/src/components/user.js
```diff
@@ -179,13 +181,13 @@ async function getUserProfile(userGuid, userName) {
...
 function parseFacilityPermissions(userResponse) {
   return userResponse.facility_permission
-    .filter((fp) => fp.ofm_portal_access)
     .map((fp) => {
       return {
         ...new MappableObjectForFront(fp, UsersPermissionsFacilityMappings).data,
         ...new MappableObjectForFront(fp.facility, UserProfileFacilityMappings).data,
       }
     })
+    .filter((fp) => fp.ofmPortalAccess && fp.programCode !== OFM_PROGRAM_CODES.CCOF)
```
weskubo-cgi, line 190: This filtering will be done in the ProviderProfile endpoint as well.

frontend/src/services/applicationService.js
```diff
@@ -257,14 +267,11 @@ export default {
...
-  async hasActiveApplicationOrFundingAgreement(facilities = []) {
+  async hasApprovedApplication(facilities = []) {
     const results = await Promise.all(
       facilities.map(async (facility) => {
-        const applications = await this.getActiveApplicationsByFacilityId(facility.facilityId)
-        return applications?.some(async (application) => {
-          application.fundingAgreement = await FundingAgreementService.getActiveFundingAgreementByApplicationId(application.applicationId)
-          return this.checkApplicationStatus(application) || this.checkFundingAgreement(application?.fundingAgreement)
-        })
+        const applications = await this.getApprovedApplicationsByFacilityId(facility.facilityId)
```
weskubo-cgi, line 273: We don't actually need the applications here...just a count so we could improve performance by not returning them all from the backend (since we're stuck getting them all from CRM). We did something similar with reportService.getDraftSurveyResponsesCountByFacility. You could also pass in a list directly and reduce the number of API calls.

weskubo-cgi (changes requested):

backend/src/components/user.js
vietle-cgi: that's correct. I add this filter just in case there's some issue in CRM (e.g.: CRM hasn't removed CCOF facilities from the ProviderProfile response yet)

frontend/src/services/applicationService.js
vietle-cgi: good point! I will update it

backend/src/components/user.js
weskubo-cgi: Great. Doesn't hurt.

frontend/src/services/applicationService.js
vietle-cgi: it's updated

weskubo-cgi (approved):

PR #389, opened 2024-10-21
OFMCC-6327 - Funding Reallocation View

vietle-cgi:


frontend/src/components/funding/FundingAllocationInfoTable.vue
```diff
@@ -0,0 +1,91 @@
...
+</template>
+
+<script>
+export default {
+  name: 'FundingAllocationInfoTable',
+}
+</script>
+
+<style scoped>
+.table {
```
weskubo-cgi, line 71: I know this is scoped but can we use a more descriptive name so that it doesn't overlap with an html element?

frontend/src/components/funding/FundingAllocationInfoTable.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <v-container fluid class="pa-0">
+    <div v-if="$vuetify.display.mdAndUp" class="table">
```
weskubo-cgi, line 3: You could have a single div with class="table" and use templates for your if/else. It would save the class duplication but add more tags. Either way is fine.

frontend/src/views/funding/FundingOverviewView.vue
```diff
@@ -12,7 +12,7 @@
           <v-icon size="large">mdi-history</v-icon>
           <strong>Payment Records</strong>
         </v-tab>
-        <v-tab v-if="hasPermission(PERMISSIONS.VIEW_FUNDING_AMOUNTS)" value="funding-allocation">
+        <v-tab v-if="hasPermission(PERMISSIONS.VIEW_FUNDING_AGREEMENT)" value="funding-allocation">
```
weskubo-cgi, line 15: Can we add a TODO to confirm permissions? Same below.

weskubo-cgi (changes requested):

frontend/src/components/funding/FundingAllocationInfoTable.vue
vietle-cgi: it's updated.

frontend/src/components/funding/FundingAllocationInfoTable.vue
vietle-cgi: it's updated.

frontend/src/views/funding/FundingOverviewView.vue
vietle-cgi: I've added TODO

weskubo-cgi (approved):

PR #422, opened 2024-11-20
Ofmcc 6475 pcm verification on funding table

jenbeckett:


frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -149,7 +149,8 @@ export default {
       }
     },
     async loadIrregularExpenses(activeFA) {
-      const expenseApplications = await IrregularExpenseService.getIrregularExpenseApplications(activeFA.applicationId, IRREGULAR_EXPENSE_STATUS_CODES.APPROVED)
+      const getPCMValidation = true
```
weskubo-cgi, line 152: I don't think this constant adds much value as it's just passed into the service call.

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -92,7 +92,7 @@ export default {
             applications?.map((application) =>
               ApplicationService.getSupplementaryApplicationsByDate(
                 application.applicationId,
-                `statusCode=${SUPPLEMENTARY_APPLICATION_STATUS_CODES.APPROVED}`,
+                `statusCode=${SUPPLEMENTARY_APPLICATION_STATUS_CODES.APPROVED}&pcmValidated=true`,
```
weskubo-cgi, line 95: I know it's existing code but I'm not a fan of passing in a generic filterQuery in our services as it allows the caller to pass in absolutely anything. Since it's a pretty minor refactor can we just pass in 4 explicit params instead (status, startDateFrom, startDateTo, pcmValidated)?

weskubo-cgi (changes requested):

frontend/src/components/funding/FundingAgreementsTab.vue
jenbeckett: I can take it out - I just thought it might be less confusing to look at instead of seeing just a random bool!

weskubo-cgi: If anything the service should document it :).

frontend/src/components/funding/FundingAgreementsTab.vue
jenbeckett: done!

weskubo-cgi (approved):

PR #428, opened 2024-12-04
Convert deprecated DCs into Deployments

trev-dev:


tools/openshift/backend.deployment.yaml
```diff
@@ -6,32 +6,31 @@ labels:
 metadata:
   name: '${REPO_NAME}-backend-dc'
```
arcshiftsolutions, line 7: Not sure if you want to change DC in the name

tools/openshift/frontend-tls.deployment.yaml
```diff
@@ -6,31 +6,33 @@ labels:
 metadata:
   name: '${REPO_NAME}-frontend-dc'
```
arcshiftsolutions, line 7: Same here - not sure if DC is ok

PR #439, opened 2025-01-10
Add cancelled and ineligble applications to application history view

trev-dev:
Jira: OFMCC-6780


trev-dev:
There's too much going on here.  Forking and cherry picking fixes.

PR #451, opened 2025-03-13
changes for decline feature

Panika-cgi:
User can decline funding agreement when they have an option to sign the funding agreement.

https://eccbc.atlassian.net/browse/OFMCC-6948

### DIFF
--- a/frontend/src/components/funding/FundingAgreementsTab.vue
+++ b/frontend/src/components/funding/FundingAgreementsTab.vue
@@ -21,7 +21,7 @@
           {{ format.formatDate(item?.endDate) }}
         </template>
         <template #[`item.statusName`]="{ item }">
-          <span :class="getStatusClass(item?.statusCode)" style="white-space: nowrap">{{ item?.statusName }}</span>
+          <span :class="getStatusClass(item?.statusCode)" class="text-no-wrap">{{ item?.statusName }}</span>
         </template>
         <template #[`item.actions`]="{ item }">
           <v-row no-gutters class="my-2 align-center justify-end justify-md-start">
--- a/frontend/src/components/funding/FundingAgreementsTab.vue
+++ b/frontend/src/components/funding/FundingAgreementsTab.vue
@@ -25,7 +25,7 @@
         </template>
         <template #[`item.actions`]="{ item }">
           <v-row no-gutters class="my-2 align-center justify-end justify-md-start">
-            <AppButton v-if="showSign(item)" :primary="false" size="small" @click="goToPDFViewer(item)">Sign</AppButton>
+            <AppButton v-if="showSign(item)" :primary="false" size="small" @click="goToPDFViewer(item)">Review for Signature</AppButton>
             <AppButton v-else-if="showOpen(item)" :primary="false" size="small" @click="goToPDFViewer(item)">Open</AppButton>
           </v-row>
         </template>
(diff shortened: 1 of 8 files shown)


frontend/src/views/funding/BaseFundingView.vue
```diff
@@ -190,6 +219,25 @@ export default {
         this.setFailureAlert('Failed to submit funding agreement', error)
       }
     },
+    async decline() {
+      const payload_decliner = {
+        ofm_provider_decliner: this.userInfo?.contactId,
```
jenbeckett, line 224: We don't generally use the CRM logical values in the frontend. Take a look at some of the mappings we do in other places. We should instead, add ofm_provider_decliner ( and date) values to the map in the backend.

frontend/src/router/index.js
```diff
@@ -174,6 +174,17 @@ const router = createRouter({
...
+    },
+    {
+      path: '/funding/declined',
+      name: 'funding-declined',
+      component: FundingConfirmationView,
+      meta: {
+        requiresAuth: true,
+        permission: PERMISSIONS.VIEW_FUNDING_AGREEMENT,
+      },
+      props: { type: 'declined' },
```
jenbeckett, line 187: Instead of passing a prop here from the router to the component - I think it would be cleaner if we just checked the name of the route on the component to verify if we are looking at the funding declined / accepted page

jenbeckett (changes requested):

frontend/src/components/funding/FundingAgreementsTab.vue
```diff
@@ -21,7 +21,7 @@
           {{ format.formatDate(item?.endDate) }}
         </template>
         <template #[`item.statusName`]="{ item }">
-          <span :class="getStatusClass(item?.statusCode)">{{ item?.statusName }}</span>
+          <span :class="getStatusClass(item?.statusCode)" style="white-space: nowrap">{{ item?.statusName }}</span>
```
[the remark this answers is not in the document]
Panika-cgi: Added.

frontend/src/views/funding/BaseFundingView.vue
Panika-cgi: thought of it and then forgot to do that. Added values to map now.

frontend/src/router/index.js
Panika-cgi: Confirmation View will use route name instead now.

frontend/src/views/funding/BaseFundingView.vue
Panika-cgi: On a second thought, we dont really need new variables - could have just used the variables used for submit. Made it to use existing variables instead.

weskubo-cgi (commented):
Please use kebab case (hyphenated) not camel case for branch names for consistency. No need to change, just for future reference.

frontend/src/router/index.js
```diff
@@ -175,6 +175,15 @@ const router = createRouter({
         permission: PERMISSIONS.VIEW_FUNDING_AGREEMENT,
       },
     },
+    {
```
weskubo-cgi, line 178: I know this has gone back and forth but I think we can do this with a single route (we can just use the existing one) that accepts a "result" param/prop bound to FundingConfirmationView. When navigating via router.push we just pass in the result as either "declined" or "accepted" and FundingConfirmationView uses these values to determine what to display. See https://router.vuejs.org/guide/essentials/passing-props.

weskubo-cgi (changes requested):

frontend/src/router/index.js
Panika-cgi: Changed to use single route.

weskubo-cgi (approved):
New approach looks great and minimizes the amount of code.

jenbeckett (approved):

PR #455, opened 2025-03-24
Added new status to Application History

Panika-cgi:
Provider declined status to FA page was already added with :https://eccbc.atlassian.net/browse/OFMCC-6948

Provider declined status added to Application history page in this PR for : https://eccbc.atlassian.net/browse/OFMCC-7070
New status of Expired added to Application history page in this PR for  : https://eccbc.atlassian.net/browse/OFMCC-7144


frontend/src/views/applications/ApplicationsHistoryView.vue
```diff
@@ -222,7 +222,7 @@ export default {
       return this.applications?.some((application) => ApplicationService.isValidApplication(application)) && this.hasGoodStanding
     },
     filteredApplicationItems() {
-      const hiddenCodes = [APPLICATION_STATUS_CODES.EXPIRED, APPLICATION_STATUS_CODES.REDIRECTED]
+      const hiddenCodes = [APPLICATION_STATUS_CODES.REDIRECTED]
```
jenbeckett, line 225: if we are only hiding redirected applications - do we need to save it into the var hiddenCodes? Or can we just use  APPLICATION_STATUS_CODES.REDIRECTED directly?

frontend/src/utils/constants.js
```diff
@@ -72,9 +72,10 @@ export const APPLICATION_STATUS_CODES = Object.freeze({
   APPROVED: 6,
   CANCELLED_BY_MINISTRY: 7, // INACTIVE state
   CANCELLED_BY_SP: 8, // INACTIVE state
-  EXPIRED: 9, // INACTIVE state
+  EXPIRED: 13, // INACTIVE state
```
jenbeckett, line 75: These should get re-ordered so they are ascending numbers.

jenbeckett (changes requested):

frontend/src/views/applications/ApplicationsHistoryView.vue
Panika-cgi: You're right! since we're only hiding redirected applications now, we could just use APPLICATION_STATUS_CODES.REDIRECTED directly.
I left it as a hiddenCodes in case we need to add more statuses later, but let me know if you think we want to use it directly.

frontend/src/utils/constants.js
Panika-cgi: re-ordered!

jenbeckett (approved):

PR #466, opened 2025-03-27
removed draft from top up query

jenbeckett:


backend/src/components/fundingAgreements.js
```diff
@@ -32,7 +33,8 @@ async function getFundingAgreements(req, res) {
     })
 
     if (req.query?.includeTopUp) {
-      const topUps = await getTopUpFundingByFilter(filter)
+      const topUpFilter = filter + ` and statuscode ne ${TOP_UP_FUNDING_STATUS_CODES.DRAFT}`
```
weskubo-cgi, line 36: It's a little odd to combine string concatenation and template literals. You can just include filter in the template literal.

weskubo-cgi (changes requested):

weskubo-cgi (approved):

PR #472, opened 2025-04-02
Added new status to Application and changed the shade of status yellow

Panika-cgi:
https://eccbc.atlassian.net/browse/OFMCC-7023

### DIFF
--- a/frontend/src/assets/css/main.css
+++ b/frontend/src/assets/css/main.css
@@ -118,7 +118,7 @@ input[type='number'] {
 }
 
 .status-yellow {
-  background-color: #ffe082;
+  background-color: #fff2b3;
   border-radius: 5px;
   padding: 2px 6px 2px 6px;
 }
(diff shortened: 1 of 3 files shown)


jenbeckett (approved):

frontend/src/assets/css/main.css
```diff
@@ -118,7 +118,7 @@ input[type='number'] {
 }
 
 .status-yellow {
-  background-color: #ffe082;
+  background-color: #ffe69a;
```
weskubo-cgi, line 121: I've spoken with Motria and we should standardize on #f9f1c6 which is already used and is the "warning" standard for BC Gov system design.

weskubo-cgi (changes requested):

frontend/src/assets/css/main.css
Panika-cgi: Changed the color as suggested, Thanks!

weskubo-cgi (approved):

PR #474, opened 2025-04-07
updated to help and resources

Panika-cgi:
Added changes to Help and resources as per : https://eccbc.atlassian.net/browse/OFMCC-6471

### DIFF
--- a/frontend/src/views/HelpView.vue
+++ b/frontend/src/views/HelpView.vue
@@ -39,9 +39,21 @@
           </a>
         </v-expansion-panel-text>
         <v-expansion-panel-text v-if="panelComponent.id === 'additional'">
+          <a href="https://www.youtube.com/watch?v=o8nPVa9xq8Q" target="_blank">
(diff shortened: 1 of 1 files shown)


frontend/src/views/HelpView.vue
```diff
@@ -74,6 +86,18 @@
           <span class="supplementary-header-label">{{ panelComponent.title }}</span>
         </v-expansion-panel-title>
         <v-expansion-panel-text v-if="panelComponent.id === 'manuals'">
+          <a href="https://www2.gov.bc.ca/assets/download/DD532A29A11B4BE1AF56FD0A1929FB26" target="_blank">
```
weskubo-cgi, line 89: I think the Template should go after Manual to be consistent with the existing 2024 links.

weskubo-cgi (changes requested):

frontend/src/views/HelpView.vue
Panika-cgi: changed the order, Thanks

weskubo-cgi (approved):

PR #479, opened 2025-04-17
Wording updates and conditional banner

Panika-cgi:
https://eccbc.atlassian.net/browse/OFMCC-6459

- The wireframes had the wrong styling, confirmed that we don't want to change the styling of the banners.
- For the  conditional banner confirmed that we want to only check FAs for user's facilities that are status active and status reason active.
- placeholder text for empty re-allocation table already matches the requirements.


frontend/src/components/funding/FundingAllocationTab.vue
```diff
@@ -51,22 +61,37 @@ export default {
       selectedFacility: null,
       fundingDetails: {},
       fundingReallocationRequests: [],
+      hasActiveFundingAgreements: true,
```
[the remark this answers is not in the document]
Panika-cgi: I am assuming you mean it should default to false?

frontend/src/components/funding/FundingAllocationTab.vue
```diff
@@ -51,22 +61,37 @@ export default {
...
     showTopupCard() {
       return Number(this.fundingDetails?.topupEnvelopeProgramming) > 0
     },
   },
 
-  created() {
+  async created() {
     this.REQUEST_CATEGORY_NAMES = REQUEST_CATEGORY_NAMES
+    const facilityList = this.userInfo?.facilities || []
+    if (facilityList.length) {
```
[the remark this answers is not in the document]
Panika-cgi: done.

frontend/src/components/funding/FundingAllocationTab.vue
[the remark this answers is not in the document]
weskubo-cgi: Yes, sorry!

frontend/src/components/funding/FundingAllocationTab.vue
```diff
@@ -51,22 +61,37 @@ export default {
...
+      await this.checkActiveFAs(facilityList)
+    }
   },
 
   methods: {
+    async checkActiveFAs(facilityList) {
+      for (const facility of facilityList) {
```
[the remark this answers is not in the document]
Panika-cgi: I was initially debating whether to add a new endpoint, but I agree this is a cleaner and more maintainable approach. I've gone ahead and incorporated all the suggested changes.

backend/src/components/fundingAgreements.js
```diff
@@ -116,11 +117,33 @@ async function getFundingReallocationRequests(req, res) {
...
 
+async function fundingAgreementExists(req, res) {
+  try {
+    const facilityIds = req.body?.facilityIds
+    const filter = facilityIds.map((id) => `(_ofm_facility_value eq ${id})`).join(' or ')
+    const operation = `ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode,ofm_version_number&$filter=${encodeURIComponent(filter)}&$top=500`
+    const response = await getOperation(operation)
+    const activeFAs = []
+    response?.value?.forEach((fa) => {
+      if (fa._ofm_facility_value && fa.statuscode === FUNDING_AGREEMENT_STATUS_CODES.ACTIVE && fa.statecode === FUNDING_AGREEMENT_STATE_CODES.ACTIVE) {
```
weskubo-cgi: Instead of filtering the statuscode and statecode after the fact you should just incorporate them into the operation filter.

backend/src/components/fundingAgreements.js
```diff
@@ -116,11 +117,33 @@ async function getFundingReallocationRequests(req, res) {
   }
 }
 
+async function fundingAgreementExists(req, res) {
+  try {
+    const facilityIds = req.body?.facilityIds
+    const filter = facilityIds.map((id) => `(_ofm_facility_value eq ${id})`).join(' or ')
+    const operation = `ofm_fundings?$select=ofm_fundingid,ofm_funding_number,ofm_declaration,ofm_start_date,ofm_end_date,_ofm_application_value,_ofm_facility_value,statuscode,statecode,ofm_version_number&$filter=${encodeURIComponent(filter)}&$top=500`
```
weskubo-cgi: We don't need to select anything besides ofm_fundingid since we're not actually returning fas.

weskubo-cgi: Is this supposed to be top or pageSize? Really we only need a single result for this to be true. So pageSize=1 would probably be most efficient.

backend/src/components/fundingAgreements.js
```diff
@@ -116,11 +117,33 @@ async function getFundingReallocationRequests(req, res) {
...
+    const response = await getOperation(operation)
+    const activeFAs = []
+    response?.value?.forEach((fa) => {
```
weskubo-cgi, line 129: I don't think you ever need to store the activeFAs. You can just check if response.value is empty (or something like that). See how userExists() works.

frontend/src/components/funding/FundingAllocationTab.vue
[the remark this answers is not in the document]
weskubo-cgi: Looks great overall. I just have a couple small comments I'll leave.

weskubo-cgi (changes requested):

backend/src/components/fundingAgreements.js
weskubo-cgi: This was addressed.

backend/src/components/fundingAgreements.js
weskubo-cgi: This was addressed.

backend/src/components/fundingAgreements.js
Panika-cgi: Right! added.

backend/src/components/fundingAgreements.js
Panika-cgi: made it to be top=1.

backend/src/components/fundingAgreements.js
Panika-cgi: matched it to userExists. :)

backend/src/components/fundingAgreements.js
Panika-cgi: back to pageSize! Turns out it broke because I gave it an extra $, Embarrassingly fixed.

weskubo-cgi: I was wondering about that...was just about to test it out for myself. Good catch!

weskubo-cgi (approved):

PR #488, opened 2025-05-01
clean up from old requirements

Panika-cgi:


jenbeckett (approved):

weskubo-cgi (approved):

frontend/src/components/applications/review/OperatingCostsSummary.vue
```diff
@@ -51,7 +51,7 @@
       </div>
 
       <!-- UPLOAD DOCUMENTS -->
-      <div v-if="showUploadedDocs" class="mt-4">
+      <div v-if="showUploadedDocs || !isRenewal" class="mt-4">
```
weskubo-cgi, line 54: It doesn't make sense to have a computed method (which is only used for this dev) and also add explicit conditions to the v-if statement. All the logic should be in the computed method.

frontend/src/views/applications/OperatingCostsView.vue
```diff
@@ -49,7 +49,7 @@
         <AppMissingInfoError v-if="showErrorMessage && isOwnedWithMortgage && !isMortgageDocsUploaded">
           {{ APPLICATION_ERROR_MESSAGES.DOCUMENT_MORTGAGE_UPLOAD }}
         </AppMissingInfoError>
-        <AppMissingInfoError v-if="showErrorMessage && isRentLease && !isRenewal && (!isFinancialDocsUploaded || !isBalanceSheetUploaded)">
+        <AppMissingInfoError v-if="showErrorMessage && !isRenewal && (!isFinancialDocsUploaded || !isBalanceSheetUploaded)">
```
weskubo-cgi, line 52: Can we convert the v-if logic on the errors to computed methods? I don't like to see more than one or two conditions on a computed method. So showRentLeaseError(), showMortgageError() etc.

weskubo-cgi (changes requested):

frontend/src/views/applications/OperatingCostsView.vue
Panika-cgi: Added 3 new computed methods

weskubo-cgi (approved):

PR #495, opened 2025-05-07
OFMCC-7154: task is completed.

infstar:
OFMCC-7154: task is completed.

### DIFF
--- a/backend/src/util/mapping/Mappings.js
+++ b/backend/src/util/mapping/Mappings.js
@@ -382,7 +382,7 @@ const PaymentMappings = [
   { back: 'ofm_payment_type', front: 'paymentTypeCode' },
   { back: 'ofm_payment_type@OData.Community.Display.V1.FormattedValue', front: 'paymentTypeName' },
   { back: 'ofm_amount', front: 'amount' },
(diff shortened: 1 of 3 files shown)


backend/src/util/mapping/Mappings.js
```diff
@@ -382,7 +382,7 @@ const PaymentMappings = [
   { back: 'ofm_payment_type', front: 'paymentTypeCode' },
   { back: 'ofm_payment_type@OData.Community.Display.V1.FormattedValue', front: 'paymentTypeName' },
   { back: 'ofm_amount', front: 'amount' },
-  { back: 'ofm_invoice_date', front: 'invoiceDate' },
+  { back: 'ofm_invoice_received_date', front: 'invoiceDate' },
```
weskubo-cgi, line 385: Given that ofm_invoice_date is still a separate field and theoretically one we might want to use in the future can we map ofm_invoice_received_date to **invoiceReceivedDate** and use that property in the frontend? The current map could lead to confusion.

weskubo-cgi (changes requested):

weskubo-cgi (approved):

PR #496, opened 2025-05-09
OFMCC-5071: task is completed.

infstar:
OFMCC-5071: task is completed.

### DIFF
--- a/frontend/src/views/HomeView.vue
+++ b/frontend/src/views/HomeView.vue
@@ -70,6 +70,7 @@
 <script>
 import { mapState } from 'pinia'
 
+import AppAlertBanner from '@/components/ui/AppAlertBanner.vue'
 import AppHeroImage from '@/components/ui/AppHeroImage.vue'
 import alertMixin from '@/mixins/alertMixin'
 import NewRequestDialog from '@/components/messages/NewRequestDialog.vue'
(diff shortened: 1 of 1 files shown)


frontend/src/views/HomeView.vue
```diff
@@ -88,13 +97,16 @@ export default {
...
-    ...mapState(useAuthStore, ['userInfo']),
+    ...mapState(useAuthStore, ['userInfo', 'isMinistryUser', 'impersonateId', 'permissions']),
     isLoading() {
       return this.loading || this.loadingFA
     },
     showReportsAlertBanner() {
       return this.hasPermission(this.PERMISSIONS.SUBMIT_DRAFT_REPORTS) && this.pendingReportsCount > 0
     },
+    isMinistryUserRequiredToImpersonate() {
+      return this.isMinistryUser && !this.impersonatedId && (!Array.isArray(this.permissions) || this.permissions.length === 0)
```
weskubo-cgi, line 108: There's a typo on **impersonatedId** that is throwing console errors. Also, was there any reason to add a permissions check? If the screen is empty because they are impersonating a user with no permissions I don't think we need to show the alert. We have a separate ticket to prevent users from impersonating an account with no permissions.

weskubo-cgi (changes requested):

frontend/src/views/HomeView.vue
```diff
@@ -18,6 +18,15 @@
         <p class="home-overview page-overview">Welcome to your Management Portal - Connect with the program, view and manage accounts, applications, and reports</p>
       </v-col>
     </v-row>
+    <v-row v-if="isMinistryUserRequiredToImpersonate">
+      <v-col class="pa-1">
+        <v-alert density="compact" type="info" variant="tonal" border="left" class="ml-2 mr-2 mt-1 mb-2">
```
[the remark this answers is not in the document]
infstar: This is done.

frontend/src/views/HomeView.vue
infstar: Oops,  thanks for the good catch on type, and I totally agree with no permission check.    Commit is done accordingly.

weskubo-cgi (approved):

PR #505, opened 2025-05-21
Ofmcc 7443 retro date supp

jenbeckett:
Ofmcc-7402 : Fix date at top of page to align with the expiry date of the Funding Agreement. 

OFMCC 7443: Add a time zone string to the date field to fix the off by 1 error seen on the retroactive date field in CRM.

### DIFF
--- a/backend/src/components/applications.js
+++ b/backend/src/components/applications.js
@@ -155,14 +155,7 @@ async function getSupplementaryApplicationById(req, res) {
 
 async function createSupplementaryApplication(req, res) {
   try {
-    const payload = new MappableObjectForBack(req.body, SupplementaryApplicationMappings).toJSON()
-    if (payload.ofm_indigenous_expenses) {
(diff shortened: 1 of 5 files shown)


backend/src/components/applications.js
```diff
@@ -192,6 +178,22 @@ async function updateSupplementaryApplication(req, res) {
...
 
+function formatSupplementaryApplication(payload) {
+  if (payload.ofm_indigenous_expenses) {
+    payload.ofm_indigenous_expenses = payload.ofm_indigenous_expenses.toString()
+  } else if (payload.ofm_needs_expenses) {
+    payload.ofm_needs_expenses = payload.ofm_needs_expenses.toString()
+  }
+
+  if (payload.ofm_retroactive_date) {
+    payload.ofm_retroactive_date = `${payload.ofm_retroactive_date}T08:00:00Z`
```
weskubo-cgi, line 189: As discussed, we should be sending the current timezone, not a modified time.

weskubo-cgi (changes requested):

